PR #8, opened 2017-06-08
CMake subdirectory/library build support

OwenDelahoy:
First I would like to thank you for your work in creating facil.io

I wanted to be able to easily use facil.io as a dependency in my project which is built with CMake, so I have added a `CMakeLists.txt`.

To use facil.io in a CMake build you may add it as a submodule to the project's repository.
`git submodule add https://github.com/boazsegev/facil.io.git`

then add the following line the project's `CMakeLists.txt` 
`add_subdirectory(facil.io)`

Here is an example of a project that is dependant on facil.io and built with CMake.
https://github.com/OwenDelahoy/facil.io-examples

However, I did have linking problems with regards to the `inline` methods and chose to make them `static inline`, does this create any unintended consequences?
https://github.com/boazsegev/facil.io/compare/master...OwenDelahoy:master#diff-6eb3b857baf5d5a8e5cc99b6ad6939d
https://github.com/boazsegev/facil.io/compare/master...OwenDelahoy:master#diff-798158459ab3baf39579beafdfb08f3d4

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,39 @@
+project(facil.io C)
+
+find_package(Threads REQUIRED)
+
+set(facil.io_SOURCES
+  src/services/pubsub.c
+  src/http/http1_simple_parser.c
+  src/http/http_request.c
+  src/http/websockets.c
+  src/http/http1_request.c
+  src/http/http_response.c
+  src/http/unused/hpack.c
+  src/http/http.c
+  src/http/http1.c
+  src/http/http1_response.c
+  src/bscrypt/bscrypt/sha1.c
+  src/bscrypt/bscrypt/hex.c
+  src/bscrypt/bscrypt/misc.c
+  src/bscrypt/bscrypt/siphash.c
+  src/bscrypt/bscrypt/random.c
+  src/bscrypt/bscrypt/base64.c
+  src/bscrypt/bscrypt/unused/mempool.c
+  src/bscrypt/bscrypt/sha2.c
+  src/bscrypt/bscrypt/xor-crypt.c
+  src/core/defer.c
+  src/core/facil.c
+  src/core/sock.c
+  src/core/evio.c
+)
+
+add_library(facil.io ${facil.io_SOURCES})
+target_link_libraries(facil.io PRIVATE Threads::Threads)
+target_include_directories(facil.io
+  PUBLIC  src/http
+  PRIVATE src/bscrypt
+  PRIVATE src/bscrypt/bscrypt
+  PUBLIC  src/core
+  PRIVATE src/core/types
+)
--- a/src/http/http.h
+++ b/src/http/http.h
@@ -155,7 +155,7 @@ A NULL terminating byte is written.
 
 Returns the number of bytes actually written (excluding the NULL byte).
 */
-inline size_t http_ul2a(char *dest, size_t num) {
+static inline size_t http_ul2a(char *dest, size_t num) {
   uint8_t digits = 1;
   size_t tmp = num;
   while ((tmp /= 10))
--- a/src/services/pubsub.c
+++ b/src/services/pubsub.c
@@ -50,10 +50,10 @@ static fio_ht_s pubsub_patterns = FIO_HASH_TABLE_STATIC(pubsub_patterns);
 
 spn_lock_i pubsub_GIL = SPN_LOCK_INIT;
 
-inline uint64_t atomic_bump(volatile uint64_t *i) {
+static inline uint64_t atomic_bump(volatile uint64_t *i) {
   return __sync_add_and_fetch(i, 1ULL);
 }
-inline uint64_t atomic_cut(volatile uint64_t *i) {
+static inline uint64_t atomic_cut(volatile uint64_t *i) {
   return __sync_sub_and_fetch(i, 1ULL);
 }
 


CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+project(facil.io C)
+
+find_package(Threads REQUIRED)
+
+set(facil.io_SOURCES
```
boazsegev: Cool!

Any way to make this list dynamic using src folders?

I think this will reduce the risk of future changes breaking the linkage.

CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+  src/bscrypt/bscrypt/xor-crypt.c
+  src/core/defer.c
+  src/core/facil.c
+  src/core/sock.c
+  src/core/evio.c
+)
+
+add_library(facil.io ${facil.io_SOURCES})
+target_link_libraries(facil.io PRIVATE Threads::Threads)
+target_include_directories(facil.io
```
boazsegev: I think we're missing `src/services` for the `pubsub` and future planned services (such as a Redis pub/sub engine and perhaps TLS/SSL support or a file caching service).

CMakeLists.txt
OwenDelahoy: Sure I will modify this to search for *.c files.

CMakeLists.txt
OwenDelahoy: Curious... There were no errors when I built it... I will double check and try to discover why.

boazsegev: The library should compile correctly with services removed or disabled... as long as you don't use the `pub/sub` service, you wouldn't know it wasn't compiled along with the rest of the files.

That's why [I sometimes use `weak` linker symbols](https://github.com/boazsegev/facil.io/blob/b4680d11e72b7e4864ec1afc611de8cdbbcddd36/src/core/facil.c#L609-L611). They allow my to gracefully fallback when some services are removed.

CMakeLists.txt
boazsegev: 🎉

CMakeLists.txt
OwenDelahoy: Ah, I see, that makes sense now.

Which headers will be needed by applications that use facil.io?

Would you consider having an include folder for applications that use facil.io?

boazsegev: > Would you consider having an include folder for applications that use facil.io?

I know it makes sense to have a separate `include` (vs. `src`) folder, but I find it confusing for development purposes... too often, the `include` folder becomes one long list of files that's very hard to navigate.

I prefer making having a [dynamic `makefile` implementation](https://github.com/boazsegev/facil.io/blob/b4680d11e72b7e4864ec1afc611de8cdbbcddd36/makefile#L52-L59) (or a dynamic CMake implementation) than making source-code navigation harder.

> Which headers will be needed by applications that use facil.io?

I guess that depends what each application does.

If the application uses the HTTP extensions, than it should include the `http.h` or the `websockets.h` files. However, these files have their own `include` directives for the `facil.h`, `http_request.h` and `http_response.h` files.

If the application uses pub/sub, the `pubsub.h` include file would also be required.

On the other hand, an "echo" application, or someone writing a custom protocol, could probably include just the `facil.h` file... which has `include` directives for `defer.h` and `sock.h`. These applications can also ignore the `http`, `services` and `bscrypt` folders or delete them, since the `core` is designed to be a stand-alone package.

The helper `fio_*` types are also great to have (they are used by the `core`, although there are other (perhaps better) alternatives to the `fio_list.h` and `fio_hash_table.h` implementations... since I favored security and ease of use over performance (i.e., the hash function implementation uses SipHash instead of CityHash or Murmur3).

CMakeLists.txt
OwenDelahoy: This is how you may find all C files automatically in CMake.

`file(GLOB_RECURSE facil.io_SOURCES LIST_DIRECTORIES false src/*.c)`

However it is not recommended by most CMake users as. Here is some explanation as to why:
https://stackoverflow.com/questions/32411963/why-is-cmake-file-glob-evil
Basically, using glob means that the user must run the command `cmake ..` to regenerate the build files after a `git pull` adds new C files. If the user doesn't realise that new files have been added they may be confused as to why a build fails to link.

I will modify the code if you do not mind the drawback.

boazsegev: @OwenDelahoy ,

This is good point. 

Let's leave the CMake as a list and I will edit the `makefile` so it recreates the `CMakeLists.txt` file  whenever I test things on my machine (I know, I will make sure it ignores `main.c`).

This will probably be the best way to automate the `CMakeLists.txt` updates and it will also solve the question of `target_include_directories`.

boazsegev:
@OwenDelahoy ,

Thank you very much 🎉👍🏻✨👏🏻👏🏻👏🏻

I'll merge this now ☺️

boazsegev:
Hi @OwenDelahoy ,

I updated the folder structure to reflect the difference between the `facil.io` library files and any development files (development files should be ignored by `facil.io` users...

... I also updated `makefile` to automatically update the  `CMakeLists.txt`.

Please make sure everything works for you and let me know if you experience any issues.

Thanks 🙏🏻 ☺️

OwenDelahoy:
I'm really glad that you've included these changes, as it will be easier to use this with other projects.

I ran into two small problems while building, I'll push the changes to my fork.

`fio_dict.h:55:52: error: unknown type name ‘uint8_t’`
I think in all instances where you have included `fio_dict.h`, `stdint.h` had been indirectly included preventing this error. But `fio_dict.c` does not include `stdint.h`.

`fio_dict.h:56:20: warning: inline function ‘fio_dict_step’ declared but never defined`
I think the lack of `static` means that this `inline` function is not visible when compiling facil.io as a library.

boazsegev:
@OwenDelahoy 

Thanks for letting me know. I didn't test the build on linux yet and I think this is the reason `stdint.h` wasn't required.

It was such a small thing, so I fixed these manually on my end.  Thanks!

OwenDelahoy:
I tested the build using the examples.
https://github.com/OwenDelahoy/facil.io-examples
Everything builds correctly.

boazsegev:
🎉👍🏻

PR #18, opened 2017-07-27
Fixed and Update README

raminfp:


lib/facil/bscrypt/README.md
```diff
@@ -34,13 +34,14 @@ Here's a simple SHA-1 computation:
...
 #include "bscrypt.h"
+#include <assert.h>
 
 int main(int argv, const char * argv[]) {
   char * result;
   sha1_s sha1 = bscrypt_sha1_init();
   bscrypt_sha1_write(&sha1, "The quick brown fox jumps over the lazy dog ", 43);
   result = bscrypt_sha1_result(&sha1);
-  assert(result == sha1.digest.str); /* locally stored on the stack */
+  assert(result == (cahr *)sha1.digest.str); /* locally stored on the stack */
```
boazsegev: Was your compiler complaining about the type for `sha1.digest.str`?

raminfp: Yes, 
```
src/main.c: In function ‘main’:
src/main.c:9:17: warning: comparison of distinct pointer types lacks a cast
   assert(result == sha1.digest.str); /* locally stored on the stack */
```

boazsegev:
Thanks 👍🏻☺️

The `include` file is a good catch. I'm surprised that the type casting is required, but I guess different compilers have different approaches.

👍🏻

PR #25, opened 2018-01-19
Fix tautological equality comparison and fallthrough warnings

64:
Fixing a few warnings that I got during compilation.

I hope you don't mind the use of `__attribute__((fallthrough))`; I had a look in other parts of the code and you seem to be using other features like `__attribute__((unused))` so I figured it was ok to do so here in order to suppress the warnings.

### DIFF
--- a/lib/facil/core/types/fiobj/fio_siphash.c
+++ b/lib/facil/core/types/fiobj/fio_siphash.c
@@ -68,25 +68,24 @@ uint64_t fio_siphash(const void *data, size_t len) {
   switch (len) { /* fallthrough is intentional */
   case 7:
     pos[6] = w8[6];
-    __attribute__((fallthrough));
+    /* fallthrough */
   case 6:
     pos[5] = w8[5];
-    __attribute__((fallthrough));
+    /* fallthrough */
   case 5:
     pos[4] = w8[4];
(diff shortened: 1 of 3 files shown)


64:
As a side note, you should consider adding `-Werror` to the CI runs just to make sure that nothing gets missed.

boazsegev:
Hi @64,

Wow, thanks!!! 🎉🙏🏻

A quick question before I merge, any reason to prefer `__attribute__((fallthrough));` instead of [a comment (`/* fallthrough */`)](https://github.com/boazsegev/facil.io/blob/50d3764a1fa9d745021e5654314e88848d0483f5/lib/facil/http/parsers/websocket_parser.h#L218)?

B.

P.S. Thanks for catching the [`fiobj_iseq` error](https://github.com/boazsegev/facil.io/pull/25/files#diff-c52c50521f608c027d2262c1c3706e21R158)!

64:
I didn't actually realise that adding a comment would suppress the warning, that's my fault. I think a comment is better since it's a little bit more portable, so I've changed it now.

boazsegev:
Thanks! 🙏🏻🎉👍🏻

boazsegev:
P.S. added the `-Werror` flag to the testing flags

PR #34, opened 2018-08-08
Update README.md

area55git:


boazsegev:
Thanks for all your work and the PR!

I'm keeping it open, but there are a few changes I want to make.

I want the Readme to be responsive (seen well on mobile), which means that the image should be adjustable.

I would rather author the width and radius using CSS. Than I will accept the PR (so you can get credit) and switch the image with code.

Can you post the gradient color details, so I could code them in?

B.

area55git:
sin titulo-1

boazsegev:
Thanks again!!! 🎉🙏🏻👍🏻

PR #35, opened 2018-08-08
added readme header and footer files

benjcal:


benjcal:
I'm adding these files to the local repo so we don't need to depend on external hosting for #34

PR #38, opened 2018-09-24
0.7.0 remastered

boazsegev:
A lot of the code was re-written and re-organized, minimizing the name space used by the core library and consolidating the core library into a two file library (`fio.h` and `fio.c`).

* Extracted the FIOBJ library from the core library, making it an add-on that could used by the extensions (such as the HTTP extension) rather than a core requirement.

* Condensed the core library and it's namespace to two files (`fio.h` and `fio.c`) - replaced all `facil_` function names with `fio_` to accommodate the new namespace.

    ...why?

    It's a choice to sacrifice ease of maintainability in favor of ease of use.

    Although two files are more difficult to maintain than 10 files (assuming it's the same amount of code)... it seems that two files are easier for other developers to copy and paste into their projects.

* Added poll support to allow for performance testing and CYGWIN compatibility. The system epoll/kqueue calls should perform better for higher loads, but now you can see for yourself.

* Timers are now in user space, allowing for more timers and less kernel dependencies.

* The `on_idle` and `on_finish` settings in `facil_run` (now `facil_start`) were removed, replaced by the more flexible `fio_state_callback_add` approach.

And more ... Read the CHANGELOG.


benjcal:
love the idea of two files! wouldn't be to hard to make it [clib](https://github.com/clibs/clib) compatible. I can look into that if you are ok with that.

I have a dumb question but, would facil could be easily used as a shared library? If so maybe we can promote around embedded linux developers. Been working quite a while with [buildroot](https://buildroot.org/) and this feels like it could be a nice fit in that ecosystem.

Awesome work @boazsegev!!

Can't wait to start testing!

boazsegev:
@benjcal ,

Thanks for the great suggestions! 🙏🏻👍🏻

As for [clib](https://github.com/clibs/clib) - it looks great. I Hope it would be a good fit, since the 2 file core library is limited to core services (network + pub/sub) and doesn't include the HTTP / WebSocket extension.

As for [buildroot](https://buildroot.org/) - wow, I didn't know this existed, looks like a wonderful project! I think facil.io could be a perfect fit, especially considering [buildroot](https://buildroot.org/) is Linux based and real-time services (pub/sub) wonderful to have on embedded devices. I hope facil.io works on embedded systems, I haven't had the chance to test it on an embedded device.

Thanks again for the ideas and the input!
🙏🏻👍🏻

PR #48, opened 2019-02-19
Not all UNIX systems support kqueue

Low-power:
Fixed by changing test macro from `__unix__` to `BSD`.

### DIFF
--- a/lib/facil/fio.c
+++ b/lib/facil/fio.c
@@ -14,6 +14,7 @@ Feel free to copy, use and enjoy according to the license provided.
 #define FIO_INCLUDE_LINKED_LIST
 #include <fio.h>
 
+#include <sys/param.h>
 #include <ctype.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


boazsegev:
Hi @Low-power and thank you for your PR! ☺️🙏🏻

I'm thankful that you exposed this issue and I agree that this needs to be fixed.

There is also the question of `poll` performance being weak when experiencing high concurrency. but this is another concern.

However, I am not convinced that this is the best way to solve this issue, especially because it adds a header file that might not exist on some systems, which is something I want to avoid as much as I can.

Also, I'm not sure all BSD systems will necessarily support kqueue.

I think I will try to solve this using tests in the makefile and using [the macros detailed here](https://sourceforge.net/p/predef/wiki/OperatingSystems/). I might not catch all cases with the macros, but I will probably catch everything in the makefile test.

Thanks!
    Bo.

PR #49, opened 2019-02-21
sendfile(2) test macros

Low-power:
The `sendfile(2)` system call seems only available on Linux, FreeBSD kernel and XNU (Darwin kernel).

This also fixes function names when `sendfile(2)` isn't available.

### DIFF
--- a/lib/facil/fio.c
+++ b/lib/facil/fio.c
@@ -2718,8 +2718,8 @@ error:
 }
 
 #else
-static int (*sock_sendfile_from_fd)(int fd, struct packet_s *packet) =
-    sock_write_from_fd;
+static int (*fio_sock_sendfile_from_fd)(int fd, struct packet_s *packet) =
+    fio_sock_write_from_fd;
 
 #endif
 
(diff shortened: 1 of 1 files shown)


boazsegev:
Hi @Low-power ,

Thank you very much for this poll-request. Specifically fixing the `fio_sock_sendfile_from_fd` fallback with a missing `sendfile`.

However, are you sure `sendfile` is only available on Linux, macOS and FreeBSD? No other OS variants provide a `sendfile` function?

Kindly,
   Bo.

Low-power:
I have an OpenBSD 6.4 i386 test VM, no `sendfile(2)` was available in this OS. For NetBSD, I can't find any man pages or codes on Internet to indicate that NetBSD supports `sendfile`.

I uses facil.io on Solaris, this OS have a `sendfile(3EXT)` function from library `libsendfile.so.1`, the arguments of this function are same as the GNU/Linux variant; but using this function requiring a library in link time, so the `makefile` must be changed to implement it.

I currently just uses the fallback version on Solaris, and since there are some more works to do to support Solaris, such as the separated socket library, I will file more pull-requests for it.

PR #50, opened 2019-02-24
Fix implicit declaration of sendfile(2)

Low-power:
If using of `sendfile(2)` is enabled by defining `USE_SENDFILE_*` from compiler option, the corresponding `#include` will be skipped. This bug is now exposed by dynamic detection of `sendfile` from makefile, in the last commit.

This also fixes a typo of macro `USE_SENDFILE_BSD`.


boazsegev:
Nice catch, thanks ☺️👍🏻🙏🏻

PR #51, opened 2019-02-25
Fix misdetect of GCC macro __has_builtin

Low-power:
There are some tests for a pre-defined macro `__has_builtin`, however this macro may has already defined from the code, to `0`, for example `lib/facil/fiobj/fiobject.h` line 38. This cause building with a GCC version that doesn't define `__has_builtin`, to fail.

### DIFF
--- a/lib/facil/fiobj/fiobject.h
+++ b/lib/facil/fiobj/fiobject.h
@@ -378,17 +378,11 @@ Atomic reference counting
   __atomic_sub_fetch(&FIOBJECT2HEAD(o)->ref, 1, __ATOMIC_SEQ_CST)
 
 /* Select the correct compiler builtin method. */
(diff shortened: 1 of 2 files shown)


boazsegev:
Thank you! 🙏🏻

☺️👍🏻🎉

boazsegev:
...However, I can't marge the poll request, because the `foo_mem.c` changes won't work.

The `fio_mem.c` file doesn't include the `fio.h` header nor does it test for GCC. In addition, it's a legacy copy of the initial facil.io memory allocator draft - it's not used during compilation and will be removed soon (hence the folder name 'legacy').

I guess I'll copy the fix to the `fiobject.h` file instead of merging.

Low-power:
OK, thanks

PR #52, opened 2019-02-26
Port to Solaris

Low-power:
Add auto-detection of non-standard fields in `struct tm`;
Add auto-detection of SystemV socket libraries;
Compatibility fix for GCC older than version 4.5;
Fix a remaining type error in `sendfile` fallback function pointer.

### DIFF
--- a/lib/facil/fio.c
+++ b/lib/facil/fio.c
@@ -2313,11 +2313,11 @@ int fio_set_non_block(int fd) {
   if (-1 == (flags = fcntl(fd, F_GETFL, 0)))
     flags = 0;
   // printf("flags initial value was %d\n", flags);
-  return fcntl(fd, F_SETFL, flags | O_NONBLOCK
 #ifdef O_CLOEXEC
-                             | O_CLOEXEC
+  return fcntl(fd, F_SETFL, flags | O_NONBLOCK | O_CLOEXEC);
+#else
+  return fcntl(fd, F_SETFL, flags | O_NONBLOCK);
 #endif
-    );
 #elif defined(FIONBIO)
   /* Otherwise, use the old way of doing it */
   static int flags = 1;
--- a/lib/facil/fio.c
+++ b/lib/facil/fio.c
@@ -2313,7 +2313,11 @@ int fio_set_non_block(int fd) {
   if (-1 == (flags = fcntl(fd, F_GETFL, 0)))
     flags = 0;
   // printf("flags initial value was %d\n", flags);
-  return fcntl(fd, F_SETFL, flags | O_NONBLOCK | O_CLOEXEC);
+  return fcntl(fd, F_SETFL, flags | O_NONBLOCK
+#ifdef O_CLOEXEC
+                             | O_CLOEXEC
+#endif
+    );
 #elif defined(FIONBIO)
   /* Otherwise, use the old way of doing it */
   static int flags = 1;
@@ -2710,7 +2714,7 @@ error:
 }
 
 #else
-static int (*fio_sock_sendfile_from_fd)(int fd, struct packet_s *packet) =
+static int (*fio_sock_sendfile_from_fd)(int fd, fio_packet_s *packet) =
     fio_sock_write_from_fd;
 
 #endif
--- a/lib/facil/http/http.c
+++ b/lib/facil/http/http.c
@@ -18,6 +18,34 @@ Feel free to copy, use and enjoy according to the license provided.
 #include <sys/types.h>
 #include <unistd.h>
 
+#ifndef HAVE_TM_TM_ZONE
+#if defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || \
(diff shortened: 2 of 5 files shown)


lib/facil/http/http.c
```diff
@@ -2073,8 +2073,12 @@ struct tm *http_gmtime(time_t timer, struct tm *tmbuf) {
   if (timer < 0)
     return gmtime_r(&timer, tmbuf);
   ssize_t a, b;
+#ifdef HAVE_TM_TM_GMTOFF
```
boazsegev: How can we add compile-time detection for these flags in cases where the `makefile` isn't used?

lib/facil/fio.c
```diff
@@ -2313,7 +2313,11 @@ int fio_set_non_block(int fd) {
   if (-1 == (flags = fcntl(fd, F_GETFL, 0)))
     flags = 0;
   // printf("flags initial value was %d\n", flags);
-  return fcntl(fd, F_SETFL, flags | O_NONBLOCK | O_CLOEXEC);
+  return fcntl(fd, F_SETFL, flags | O_NONBLOCK
```
boazsegev: This isn't very readable... maybe repeat the whole line in the different `#if` sections rather then just repeat the flag?

lib/facil/http/http.h
```diff
@@ -7,6 +7,12 @@ Feel free to copy, use and enjoy according to the license provided.
 */
 #define H_HTTP_H
 
+#ifdef __GNUC__
```
boazsegev: What if `__GNCU_C` isn't defined?

Also...

Should this be here: https://github.com/boazsegev/facil.io/blob/master/lib/facil/fio.h#L221-L229

lib/facil/http/http.c
Low-power: I can't find a macro in `sys/time.h` that specifically indicate those additional fields are available, from either FreeBSD or GNU (GLIBC).

The GLIBC header enables those fields based on macro `__GNU_MISC`, I think it was controlled by `_GNU_SOURCE`.
While the FreeBSD header have those fields unconditionally.

lib/facil/http/http.h
Low-power: Yes

lib/facil/http/http.c
Low-power: Just seen `sys/time.h` in OS X and OpenBSD; it seems that all BSDs has those fields. 
ctime(3) for GLIBC said those 2 fields will be available if `_BSD_SOURCE` is defined.

I can add test macros for all known BSD systems, but testing on GNU system is somewhat harder; because in my test it will also be available based on some other macros, such as `_SVID_SOURCE`, in this case `_BSD_SOURCE` won't be defined by the GLIBC headers.

Low-power: I decide to test macro `_BSD_SOURCE` on GNU systems, just based on what its man page said.

boazsegev: I doubt if `_BSD_SOURCE` is defined on all BSD systems. it's a `glibc` flag often set by the developer on Linux systems.

How about testing for either `__GLIBC__` or and of the Marcos [here](https://github.com/boazsegev/facil.io/blob/ed5998968bfb8d1423a9597d758316cd9ad42fe7/lib/facil/fio.c#L47-L48) as suggested [in this SO answer](https://stackoverflow.com/a/45107992/4025095)...?

lib/facil/http/http.c
```diff
@@ -18,6 +18,34 @@ Feel free to copy, use and enjoy according to the license provided.
...
 #include <unistd.h>
 
+#ifndef HAVE_TM_TM_ZONE
+#if defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || \
+    defined(__DragonFly__) || defined(__bsdi__) || defined(__ultrix) ||	\
+    (defined(__APPLE__) && defined(__MACH__)) || \
+    (defined(__sun) && !defined(__SVR4))
+/* Known BSD systems */
+#define HAVE_TM_TM_ZONE 1
+#elif defined(__GLIBC__) && defined(_BSD_SOURCE)
```
boazsegev: I see, nicely done.

lib/facil/http/http.h
boazsegev: I would be happy if you could move this test to `fio.h` before I merge.

Either way - thank you very much for all the wonderful effort!

👏🏻👏🏻👏🏻👏🏻🙏🏻🙏🏻🙏🏻👍🏻👍🏻☺️

Low-power:
It this change ok?

Low-power:
Pushed a late bug fix!

boazsegev:
@Low-power ,

Thanks for the updates!

I need to review some of the logic regarding the macros, but it's already an improvement, so I think I'll merge the PR as is.

Side-note:  as far as I know, `-DFOO` is the same as `-DFOO=1` in the compilation flags.

Thanks again!
    Bo.

Low-power:
I didn't realized that! I previously thought that `-D FOO` was equal to `-D FOO=`

PR #53, opened 2019-02-26
Fix format warnings of type 'pid_t'

Low-power:
Solaris has defined type `pid_t` as `long int` on 32-bit platforms, so using `%d` in a format string for `pid_t` will cause a compiler warning about the type mismatch.


boazsegev:
Interesting! Thanks for the PR 👏🏻☺️🙏🏻

Question... wouldn't `long int` have a higher range? how high do Solaris process number go?

Maybe we should change the printout upwards (use `%zd` and cast all `pid_t` to `ssize_t`)...?

Low-power:
Nope, `long int` has same size with `int` on Solaris i386 (32-bit).

PR #54, opened 2019-02-26
Fix a possible unreleased lock in function block_new

Low-power:
When that allocation failed.


boazsegev:
Nice catch!
☺️🙏🏻🙏🏻👍🏻👏🏻👏🏻👏🏻

PR #55, opened 2019-02-26
Fix fio_atomic_xchange fallback implementation for older GCC

Low-power:
An `or` operation can't exchange those 2 values as expected in `fio_atomic_xchange`, makes `fio_unlock` no effect.

### DIFF
--- a/lib/facil/fio.h
+++ b/lib/facil/fio.h
@@ -2022,7 +2022,7 @@ int fio_pubsub_is_attached(fio_pubsub_engine_s *engine);
 /* Select the correct compiler builtin method. */
 #elif __has_builtin(__sync_add_and_fetch)
 /** An atomic exchange operation, ruturns previous value */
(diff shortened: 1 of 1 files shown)


Low-power:
Although `nand` operation can't really exchange any values,  but `0` and `1`, this is enough to fix the `fio_unlock` bug, since it only use `0` and `1`.

Low-power:
I force pushed a new commit that uses `__sync_val_compare_and_swap` instead.

Low-power:
Sorry, the function names was missing a `_` character in my last commit.

boazsegev:
Thank you for the PR!
☺️👍🏻🎉👏🏻

boazsegev:
I should probably find a way to test on a wider range of compilers, as well as testing with and without the `makefile` (since this is a [possible use-case](https://github.com/boazsegev/facil.io#adding-facilio-to-an-existing-project)).

Thank you very much for exposing this 🙏🏻☺️

PR #64, opened 2019-06-22
Adding Meson support for facil.io framework.

squidfarts:
Hello. 
I happen to come across your web framework and I thought it would be a nice fit so I wrote a simple meson.build file to add support for Meson build system.


boazsegev:
Duplicate. See #66

PR #66, opened 2019-06-23
Adding support for Meson build system.

squidfarts:
Hello.
I happen to come across your web framework and I thought it would be a fit so I wrote a simple Meson build script.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,108 @@
+###################################################################################
+#                                                                                 #
+# NAME: meson.build                                                               #
+#                                                                                 #
+# AUTHOR: Michael Brockus.                                                        #
+#                                                                                 #
+# CONTACT: <mailto:michael@squidfarts.com>                                        #
+#                                                                                 #
+# NOTICES:                                                                        #
+#                                                                                 #
+# MIT License                                                                     #
+#                                                                                 #
+# Copyright (c) 2019 Team Squidfarts                                              #
+#                                                                                 #
+# Permission is hereby granted, free of charge, to any person obtaining a copy    #
+# of this software and associated documentation files (the "Software"), to deal   #
+# in the Software without restriction, including without limitation the rights    #
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell       #
+# copies of the Software, and to permit persons to whom the Software is           #
+# furnished to do so, subject to the following conditions:                        #
+#                                                                                 #
+# The above copyright notice and this permission notice shall be included in all  #
+# copies or substantial portions of the Software.                                 #
+#                                                                                 #
+# DISCLAIMERS:                                                                    #
+#                                                                                 #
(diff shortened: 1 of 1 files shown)


boazsegev:
Hi @squidfarts ,

Thank you very much for your time and the PR!

It would be great to support more build systems. However, I do have some questions before we move forwards:

- Is it possible to automate the file names used for the build system? Doing so will allow for easier maintenance across changes to the system. See, for example, the [automated CMake support inn the makefile](https://github.com/boazsegev/facil.io/blob/fe7574d70c0bd3aa73a8d592cb902cfd6b8b8837/makefile#L695-L725).

- Is it possible to add auto-detection for features and adjust compilation flags according to the detected features? For example, [this is used to detect OpenSSL and define the `HAVE_OPENSSL` flag](https://github.com/boazsegev/facil.io/blob/fe7574d70c0bd3aa73a8d592cb902cfd6b8b8837/makefile#L367-L439).

If it's not possible, it okay, we'll do what we can.

squidfarts:
It is posable to automate file names in Meson.  I haven't done it before but I think it gos something like this:

cc = meson.get_compiler('c')
cdata = configuration_data()
check_headers = [
]

foreach h : check_headers
  if cc.has_header(h.get(1))
    cdata.set(h.get(0), 1)
  endif
endforeach

check_functions = [
# check token ('HAVE_TM_TM_ZONE=1',)
# check token ('HAVE_BEARSSL',)
# check token ('HAVE_BEARSSL',)
# check token ('HAVE_OPENSSL',)
# check token ('HAVE_S2N',)
# check token ('HAVE_ZLIB',)
# check token ('HAVE_POSTGRESQL',)
# check token ('HAVE_POSTGRESQL',)
]

foreach f : check_functions
  if cc.has_function(f.get(1), prefix : f.get(2))
    cdata.set(f.get(0), 1)
  endif
endforeach

configure_file(input : 'config.h.meson',
  output : 'config.h',
  configuration : cdata)


I believe it's possible to auto-detect compiler flags as well. My guess is with "if else" statements or some of Mesons build-in options.

squidfarts:
Anything else?

boazsegev:
@squidfarts ,

Sorry for the belated response. I'm taking a composition master class and our concert is tomorrow night, so my time is very limited.

As for the PR, there isn't necessarily anything else, but I don't think I can merge the PR as is... at least not yet.

There are two possible approaches I would prefer to review:

1. Use `make` to dynamically author a `meson.build` file with all of the appropriate compilation flags.

2. Author a dynamic `meson.build` file as part of this PR (which would be merged into the 0.8 development branch).

   This assumes that dynamic file names and dynamic feature / library detection are available with Meson, and could be incorporated it into the PR. The feature / library detection will need to set preprocessor compilation flags (facil.io doesn't use a `config.h` file).

Anyway, I'll look over things again after the concert.

Kindly,
    Bo.

PR #67, opened 2019-06-27
Meson support with 'facial.io'...

squidfarts:
I wrote a meson.build and some sub meson.build scripts to better modularize the Meson code and to avoid longe path variables when trying to find header files in the directory include paths.  This is currently a working progress 

Meson can dedicated libraries on the users system [here](https://mesonbuild.com/Dependencies.html).  There is also this wrap file tool for third party dependencies [here](https://mesonbuild.com/Wrap-dependency-system-manual.html).

Command line results:

`
Last login: Wed Jun 26 20:02:51 on ttys000
Michaels-MacBook-Pro:~ shnitzel$ cd Developer/github/fork/facil.io/
Michaels-MacBook-Pro:facil.io shnitzel$ meson setup build
The Meson build system
Version: 0.51.999
Source dir: /Users/shnitzel/Developer/github/fork/facil.io
Build dir: /Users/shnitzel/Developer/github/fork/facil.io/build
Build type: native build
Project name: facil.io
Project version: 0.2.4
C compiler for the build machine: cc (clang 10.0.1 "Apple LLVM version 10.0.1 (clang-1001.0.46.4)")
C compiler for the host machine: cc (clang 10.0.1 "Apple LLVM version 10.0.1 (clang-1001.0.46.4)")
Build machine cpu family: x86_64
Build machine cpu: x86_64
Run-time dependency threads found: YES 
Library m found: YES
Build targets in project: 1
Found ninja-1.9.0.git at /opt/local/bin/ninja
Michaels-MacBook-Pro:facil.io shnitzel$ ninja -C build/
ninja: Entering directory `build/'
[20/20] Linking static target libfacil.io.a.
Michaels-MacBook-Pro:facil.io shnitzel$ 
`

PS I think I may have fixed a bug in 'http1_parser.c' this error was printing to the consul until I made a fix.  Also I tried to add compiler flags and more errors printed so there is going to be a lot work to do.

`
Michaels-MacBook-Pro:facil.io shnitzel$ ninja -C build/
ninja: Entering directory `build/'
[13/20] Compiling C object 'facil.io@s..._facil_http_parsers_http1_parser.c.o'.
FAILED: facil.io@sha/lib_facil_http_parsers_http1_parser.c.o 
cc -Ifacil.io@sha -I. -I.. -Ilib -I../lib -Ilib/facil -I../lib/facil -I../lib/facil/tls -I../lib/facil/fiobj -I../lib/facil/cli -I../lib/facil/http -I../lib/facil/http/parsers -I../lib/facil/redis -flto -Xclang -fcolor-diagnostics -fsanitize=undefined -pipe -Wall -Winvalid-pch -Wextra -Wpedantic -Werror -O3 -g -MD -MQ 'facil.io@sha/lib_facil_http_parsers_http1_parser.c.o' -MF 'facil.io@sha/lib_facil_http_parsers_http1_parser.c.o.d' -o 'facil.io@sha/lib_facil_http_parsers_http1_parser.c.o' -c ../lib/facil/http/parsers/http1_parser.c
../lib/facil/http/parsers/http1_parser.c:448:66: error: missing field 'reserved' initializer [-Werror,-Wmissing-field-initializers]
        (struct http1_parser_protected_read_only_state_s){0, 0, 0};
                                                                 ^
../lib/facil/http/parsers/http1_parser.c:454:64: error: missing field 'reserved' initializer [-Werror,-Wmissing-field-initializers]
      (struct http1_parser_protected_read_only_state_s){0, 0, 0};
                                                               ^
2 errors generated.
[18/20] Compiling C object 'facil.io@sha/lib_facil_fio.c.o'.
ninja: build stopped: subcommand failed.
`

### DIFF
--- a/lib/facil/http/parsers/http1_parser.c
+++ b/lib/facil/http/parsers/http1_parser.c
@@ -445,13 +445,13 @@ re_eval:
                                               : args->on_request)(args->parser))
       goto error;
     args->parser->state =
-        (struct http1_parser_protected_read_only_state_s){0, 0, 0};
+        (struct http1_parser_protected_read_only_state_s){0, 0, 0, 0};
   }
   return CONSUMED;
 error:
   args->on_error(args->parser);
   args->parser->state =
-      (struct http1_parser_protected_read_only_state_s){0, 0, 0};
+      (struct http1_parser_protected_read_only_state_s){0, 0, 0, 0};
   return args->length;
 }
 
--- a/lib/facil/meson.build
+++ b/lib/facil/meson.build
@@ -4,7 +4,7 @@
 #                                                                                 #
 # AUTHOR: Boaz Segev.                                                             #
 #                                                                                 #
-# WRITTEN BY: Michael Brockus.                                                     #
+# WRITTEN BY: Michael Brockus.                                                    #
 #                                                                                 #
 # CONTACT: <mailto:michael@squidfarts.com>                                        #
 #                                                                                 #
--- a/lib/facil/meson.build
+++ b/lib/facil/meson.build
@@ -1,0 +1,71 @@
+###################################################################################
+#                                                                                 #
+# NAME: meson.build                                                               #
+#                                                                                 #
+# AUTHOR: Michael Brockus.                                                        #
+#                                                                                 #
+# CONTACT: <mailto:michael@squidfarts.com>                                        #
+#                                                                                 #
+# NOTICES:                                                                        #
+#                                                                                 #
+# MIT License                                                                     #
+#                                                                                 #
+# Copyright (c) 2019 Team Squidfarts                                              #
+#                                                                                 #
+# Permission is hereby granted, free of charge, to any person obtaining a copy    #
+# of this software and associated documentation files (the "Software"), to deal   #
+# in the Software without restriction, including without limitation the rights    #
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell       #
+# copies of the Software, and to permit persons to whom the Software is           #
+# furnished to do so, subject to the following conditions:                        #
+#                                                                                 #
+# The above copyright notice and this permission notice shall be included in all  #
+# copies or substantial portions of the Software.                                 #
+#                                                                                 #
+# DISCLAIMERS:                                                                    #
+#                                                                                 #
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR      #
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,        #
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE     #
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER          #
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,   #
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE   #
+# SOFTWARE.                                                                       #
+#                                                                                 #
+###################################################################################
+
+
+
+facil_src = files(
+    [
+        join_paths('fio.c'),
+        join_paths('tls', 'fio_tls_missing.c'),
+        join_paths('tls', 'fio_tls_openssl.c'),
+        join_paths('fiobj', 'fio_siphash.c'),
+        join_paths('fiobj', 'fiobj_ary.c'),
+        join_paths('fiobj', 'fiobj_data.c'),
+        join_paths('fiobj', 'fiobj_hash.c'),
+        join_paths('fiobj', 'fiobj_json.c'),
+        join_paths('fiobj', 'fiobj_mustache.c'),
+        join_paths('fiobj', 'fiobj_numbers.c'),
+        join_paths('fiobj', 'fiobj_str.c'),
+        join_paths('fiobj', 'fiobject.c'),
+        join_paths('cli', 'fio_cli.c'),
+        join_paths('http', 'http.c'),
+        join_paths('http', 'http1.c'),
+        join_paths('http', 'http_internal.c'),
+        join_paths('http', 'websockets.c'),
+        join_paths('http', 'parsers', 'http1_parser.c'),
+        join_paths( 'redis', 'redis_engine.c'),
+    ])
+
+facil_dir = include_directories(
+    [
+        '.', 
+        'tls', 
+        'fiobj', 
+        'cli', 
+        'http', 
+        'redis',
+        join_paths('http', 'parsers')
+    ])
--- a/lib/facil/meson.build
+++ b/lib/facil/meson.build
@@ -2,7 +2,9 @@
 #                                                                                 #
 # NAME: meson.build                                                               #
 #                                                                                 #
-# AUTHOR: Michael Brockus.                                                        #
+# AUTHOR: Boaz Segev.                                                             #
+#                                                                                 #
+# WRITTEN BY: Michael Brockus.                                                     #
 #                                                                                 #
 # CONTACT: <mailto:michael@squidfarts.com>                                        #
 #                                                                                 #
--- a/lib/facil/meson.build
+++ b/lib/facil/meson.build
@@ -12,7 +12,7 @@
 #                                                                                 #
 # MIT License                                                                     #
 #                                                                                 #
-# Copyright (c) 2019 Team Squidfarts                                              #
+# Copyright (c) Boaz Segev, 2018-2019                                             #
 #                                                                                 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy    #
 # of this software and associated documentation files (the "Software"), to deal   #
--- a/lib/facil/meson.build
+++ b/lib/facil/meson.build
@@ -1,38 +1,12 @@
 ###################################################################################
 #                                                                                 #
-# NAME: meson.build                                                               #
+# NAME: lib/facil/meson.build                                                     #
 #                                                                                 #
 # AUTHOR: Boaz Segev.                                                             #
 #                                                                                 #
 # WRITTEN BY: Michael Brockus.                                                    #
 #                                                                                 #
-# CONTACT: <mailto:michael@squidfarts.com>                                        #
-#                                                                                 #
-# NOTICES:                                                                        #
-#                                                                                 #
-# MIT License                                                                     #
-#                                                                                 #
-# Copyright (c) Boaz Segev, 2018-2019                                             #
-#                                                                                 #
-# Permission is hereby granted, free of charge, to any person obtaining a copy    #
-# of this software and associated documentation files (the "Software"), to deal   #
-# in the Software without restriction, including without limitation the rights    #
-# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell       #
-# copies of the Software, and to permit persons to whom the Software is           #
-# furnished to do so, subject to the following conditions:                        #
-#                                                                                 #
-# The above copyright notice and this permission notice shall be included in all  #
-# copies or substantial portions of the Software.                                 #
-#                                                                                 #
-# DISCLAIMERS:                                                                    #
-#                                                                                 #
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR      #
-# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,        #
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE     #
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER          #
-# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,   #
-# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE   #
-# SOFTWARE.                                                                       #
+# License: MIT                                                                    #
 #                                                                                 #
 ###################################################################################
 
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -13,3 +13,9 @@
 
 
 subdir('facil')
+
+facil_lib = library(meson.project_name(), 
+    sources: facil_src,
+    include_directories: facil_dir,
+    dependencies: [ threads_dep, m_dep ],
+    install: true)
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -1,38 +1,12 @@
 ###################################################################################
 #                                                                                 #
-# NAME: meson.build                                                               #
+# NAME: lib/meson.build                                                           #
 #                                                                                 #
 # AUTHOR: Boaz Segev.                                                             #
 #                                                                                 #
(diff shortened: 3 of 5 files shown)


lib/facil/http/parsers/http1_parser.c
```diff
@@ -445,13 +445,13 @@ size_t http1_fio_parser_fn(struct http1_fio_parser_args_s *args) {
                                               : args->on_request)(args->parser))
       goto error;
     args->parser->state =
-        (struct http1_parser_protected_read_only_state_s){0, 0, 0};
+        (struct http1_parser_protected_read_only_state_s){0, 0, 0, 0};
```
boazsegev: Unless compiling C++ code, this isn't necessary. Unspecified fields are automatically initialized to zero if missing (which is also true for the 2 zeros I wrote after the first zero).

lib/facil/http/parsers/http1_parser.c
```diff
@@ -445,13 +445,13 @@ size_t http1_fio_parser_fn(struct http1_fio_parser_args_s *args) {
...
   }
   return CONSUMED;
 error:
   args->on_error(args->parser);
   args->parser->state =
-      (struct http1_parser_protected_read_only_state_s){0, 0, 0};
+      (struct http1_parser_protected_read_only_state_s){0, 0, 0, 0};
```
boazsegev: Same.

lib/facil/meson.build
```diff
@@ -0,0 +1,71 @@
+###################################################################################
```
boazsegev: Please see the [contribution guide and authorship/copyright attribution](https://github.com/boazsegev/facil.io/blob/master/CONTRIBUTING.md#license).

Copyrights and attribution should be made under my name, though it may list you as "Contributed by ..." or "Written by ...". The license shouldn't be pasted in the file (although [the file may contain a short header](https://github.com/boazsegev/facil.io/blob/fe7574d70c0bd3aa73a8d592cb902cfd6b8b8837/lib/facil/fio.h#L1-L6)).

If you don't agree to the contribution terms, let me know. Basically, I just want to avoid the mess OpenSSL had to go through when updating their license and I want to maintain control over the repo.

lib/facil/meson.build
```diff
@@ -0,0 +1,71 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER          #
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,   #
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE   #
+# SOFTWARE.                                                                       #
+#                                                                                 #
+###################################################################################
+
+
+
+facil_src = files(
```
boazsegev: I see file names are still explicit...?

If this is the only option, perhaps it would be better if I automate the `mason.build` authorship in the `makefile`?

lib/meson.build
```diff
@@ -0,0 +1,39 @@
+###################################################################################
```
boazsegev: As for license - see my previous comment.

As for the file itself - do they have to be in every subfolder, or could we settle for a single `mason.build` file in the project's root (or the library's root)?

meson.build
```diff
@@ -0,0 +1,86 @@
+###################################################################################
```
boazsegev: See previous comments for license and file location / necessity.

meson.build
```diff
@@ -0,0 +1,86 @@
...
+
+
+project('facil.io', 'c',
+    version         : '0.2.4',
+    license         : 'MIT',
+    meson_version   : '>=0.50.0',
+    default_options : 
+    [ 
+        'werror=true',
+        'optimization=3',
```
boazsegev: Isn't optimization level 3 sometimes buggy?

facil.io should be okay with this level of optimization, but I would avoid it for most projects... what do you think?

meson.build
```diff
@@ -0,0 +1,86 @@
...
+        'b_lundef=true' 
+    ]
+)
+cc = meson.get_compiler('c')
+args_for_langs = 'c'
+
+
+
+threads_dep = dependency('threads')
+m_dep = cc.find_library('m', required : false)
```
boazsegev: Could we do something similar to his for OpenSSL?

In pseudo code (probably not valid mason):

```
m_dep = cc.find_library('crypt', required : false, defines: "HAVE_OPENSSL")
```

...?

meson.build
```diff
@@ -0,0 +1,86 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,   #
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE   #
+# SOFTWARE.                                                                       #
+#                                                                                 #
+###################################################################################
+
+
+
+project('facil.io', 'c',
+    version         : '0.2.4',
```
boazsegev: Is this a facil.io version specifier? 

If so, could it be dynamically loaded? How?

squidfarts: That is supposed to be the facil.io version number, if I was off then feel free to charge it to the correct number.  I don’t think there is a dynamic way to pre-load the version.

boazsegev: Cool 👍🏻

FYI: the [current version is 0.7.2](https://github.com/boazsegev/facil.io/blob/fe7574d70c0bd3aa73a8d592cb902cfd6b8b8837/lib/facil/fio.h#L110-L113). In [0.8.x the version number will appear here](https://github.com/boazsegev/facil.io/blob/c35d19002c20f88696cb67224d4cb26a349290a5/lib/facil/fio-stl.h#L165-L168).

lib/facil/http/parsers/http1_parser.c
boazsegev: I just noticed that on a high enough warning level, you get the `missing field` warning (turned into an error). This is an informative warning, but it's probably worth avoiding.

meson.build
boazsegev: > I don’t think there is a dynamic way to pre-load the version.

Hmm... not very comfortable... I wouldn't want to have the version number appear more than once, since it would be easy to make errors when maintaining the project.

Maybe we I could write a small script that will compile facil.io and bring the version to stdout, allowing us to "pipe" the version into the mason.build file during updates.

I'll have to think about this.

meson.build
squidfarts: I believe we can do something similar.  I have to run some experiments when I get back.

meson.build
squidfarts: I can remove it if it’s too much.

boazsegev: I would recommend `O2`, [that's what I set in the `makefile`](https://github.com/boazsegev/facil.io/blob/fe7574d70c0bd3aa73a8d592cb902cfd6b8b8837/makefile#L65).

lib/meson.build
squidfarts: Sorry about that, I wrote a Meson file generation tool that would scan a project and add meson.build files depending on structure, there was only supposed to be short comments with a path to the file starting from root in the sub files.

lib/facil/meson.build
squidfarts: There is no way to automate the files names at this time.

meson.build
squidfarts: Meson has a method for the project version number.  This can be done by calling ‘meson.project_verson()’.

meson.build
squidfarts: Sure thing.

lib/meson.build
squidfarts: I will also fix that.

lib/facil/http/parsers/http1_parser.c
squidfarts: I don’t think ignoring the warnings is such good idea....

boazsegev: You're right, it's worth avoiding the warning.

lib/facil/meson.build
boazsegev: Okay, thanks.

meson.build
boazsegev: If the approach you're referencing is the one [suggested here](https://github.com/mesonbuild/meson/issues/4144), then I'm against it.

Using `meson.project_verson()` to export the version number to a compilation flag makes it hard to support semantical versioning and makes it impossible to support multiple build systems in a single project.

Both make and CMake are currently supported in addition to the future mason support. I won't make mason exclusive because it requires a version number.

squidfarts: Version number is not required to define the project, I just added that so the user may use a wrap file to download the preferred version of facil.io.  Should I remove it?

lib/meson.build
squidfarts: I think I fixed all of the the licensing information.

meson.build
squidfarts: Optimization level is now set to two.

meson.build
squidfarts: How about this block of Meson script?

# meson.build experiment for OpenSSL.
`
# *** Checks for OpenSSL    ***
openssl_option = get_option('openssl')
if openssl_option.disabled()
  openssl_dep = []
else
  # XXX: https://github.com/mesonbuild/meson/issues/2945
  openssl_dep = dependency('openssl', required: false)
  if openssl_dep.found()
    backends += ['openssl']
  else
    # MSVC builds of OpenSSL does not generate pkg-config files,
    # so we check for it manually here in this case, if we can't find those files
    # Based on the CMake check for OpenSSL in CURL's CMakeLists.txt,
    # on which headers we should check for
    openssl_headers = []
    foreach h : ['crypto.h', 'engine.h', 'err.h', 'pem.h',
                 'rsa.h', 'ssl.h', 'x509.h', 'rand.h', 'tls1.h']
      openssl_headers += 'openssl/' + h
    endforeach

    # OpenSSL 1.1.x and 1.0.x (or earlier) have different .lib names,
    # so we need to look for the correct pair

    # Find either libcrypto.lib (1.1.x) or libeay32.lib (1.0.x or earlier) first
    libcrypto_dep = cc.find_library('crypto', required: false)
    if libcrypto_dep.found()
      libssl = 'ssl'
    else
      libcrypto_dep = cc.find_library('eay32', required: openssl_option)
      libssl = 'ssleay32'
    endif

    if libcrypto_dep.found()
      # Find the corresponding SSL library depending on which crypto .lib we found
      libssl_dep = cc.find_library(libssl, required: openssl_option, has_headers: openssl_headers)
    endif

    if libcrypto_dep.found() and libssl_dep.found()
      openssl_dep = [libcrypto_dep, libssl_dep]
      backends += ['openssl']
    endif
  endif
endif
`

# meson_options.txt

`option('openssl', type: 'feature', value: 'disabled', description: 'support for OpenSSL networking configration')`

lib/meson.build
boazsegev: I would still prefer if the license text was removed. Simply use:

```
License: MIT
```

No need to include the license text.

If you don't have time for this, no worries, I can clean it up after the merge.

squidfarts: Got it making one more edit.

meson.build
boazsegev: Maybe removing the version number would make more sense... yes.

I'm just not sure I understand, how does providing a version number help allows a developer to download the correct version? Right now, this is managed through git tags / releases. But I don't know that mason has a download manager...?

meson.build
boazsegev: Looks interesting but also very complex. Maybe we should drop the OpenSSL support for now and manage it the same way we will manage file names (maybe using `make` to auto-generate a new mason.build file)...?

Note that facil.io supports OpenSSL 1.1.x or higher. Lower versions aren't supported. This might cleanup some of the suggested script.

lib/meson.build
squidfarts: Ok I switched the long verbose version of the license info to use this suggested one instead.

meson.build
squidfarts: I think this can be a lot shorter if I used Meson's CMake module [here](https://mesonbuild.com/CMake-module.html).

This version of [OpenSSL](https://github.com/janbar/openssl-cmake/blob/master/CMakeLists.txt) has a CMakeLists.txt. If it's ok with you then I can add less Meson code than I presented in my latest experiment.

Just a thought.

meson.build
squidfarts: Meson has this [WrapDB](https://mesonbuild.com/Wrap-dependency-system-manual.html) system for downloading third party dependencies.
With GitHub a git-wrap file would do the trick for automatically downloading 'facial.io'.

squidfarts: Exposure of version is removed.

meson.build
squidfarts: This would result with something like this for example.

`cmake_mod = import('cmake')

#Configure the CMake project
sub_proj = cmake_mod.subproject('libsimple_cmake')

#Fetch the dependency object
cm_lib = sub_proj.dependency('cm_lib')

executable(exe1, ['sources'], dependencies: [cm_lib])`

Just like that.

boazsegev: I trust you, since I know very little about Mason, but I want to make sure that we're not forcing Mason users to install CMake.

Does this change requires people to install cmake, or does mason parses OpenSSL's CMake file?

squidfarts: Sadly yes, the CMake module route would require that the user would have to install CMake.  That’s why I don’t use it much.

squidfarts: I think there may be a way to simply the previous experiment.  I going to try some things.

squidfarts: This maybe better to use for OpenSSL support, what do you think?

`# MSVC builds of OpenSSL does not generate pkg-config files,
# so we check for it manually here in this case, if we can't find those files
openssl = dependency('openssl', required: cc.get_id() != 'msvc')

if not openssl.found()
  # OpenSSL 1.1.x and 1.0.x (or earlier) have different .lib names,
  # so we need to look for the correct pair

  # Find either libcrypto.lib (1.1.x) or libeay32.lib (1.0.x or earlier) first
  libcrypto = cc.find_library('libcrypto', required: false)
  if not libcrypto.found()
    libeay = cc.find_library('libeay32')
  endif

  # Find the corresponding SSL library depending on which crypto .lib we found
  if libcrypto.found()
    libssl = cc.find_library('libssl')
  elif libeay.found()
    ssleay = cc.find_library('ssleay32')
  endif

  # Now set the openssl dependency to the corresponding pair of .lib files that we found
  if libcrypto.found()
    openssl = [libcrypto, libssl]
  else
    openssl = [libeay, ssleay]
  endif
endif`

squidfarts: Would that work?

meson.build
squidfarts: Anything on that script yet?

meson.build
boazsegev: Sorry for the delay, my grandmother's in the hospital, so there's a bit of a time shortage.

The script looks great, though I didn't test it on my machine.

However, facil.io supports only OpenSSL 1.1.x or higher, so there's no need for the `libeay` part.

Lets fix that up and merge. I'll work on the rest of it later.

meson.build
boazsegev: I'll write it later. Let's leave out the version for now and fix things up as we go.

squidfarts: Version number removed.

meson.build
squidfarts: If we drop support for versions less than 1.1.x, the resulting script would look something like this.

```
#MSVC builds of OpenSSL does not generate pkg-config files,
#so we check for it manually here in this case, if we can't find those files
openssl = dependency('openssl', required: cc.get_id() != 'msvc')

if not openssl.found()
        libcrypto = cc.find_library('libcrypto', required: false)
        libssl = cc.find_library('libssl')
        openssl = [libcrypto, libssl]
endif
```

squidfarts: Sorry about your grandmother.

squidfarts: Well OpenSSL is added.

boazsegev:
Did you change your mind about the PR? or did you close this by mistake?

squidfarts:
By mistake.

PR #71, opened 2019-08-27
Fix memory leak `tls->trust` + various linking-related fixes

fbrausse:
Hi,

I'm very happy that you put so much effort into this library, I found it easy to use for a Websocket client application I'm building which should run on Linux and Mac OS X, thank you!

However, I encountered a few things I needed to change:
- https://github.com/boazsegev/facil.io/commit/fcde3f074ba1b35d3b69ebb5270b52686fe26d4e fixes the memory leak when `tls->trust` is not cleaned up in `fio_tls_destroy(tls)`
- Building an OpenSSL-enabled facil.io on Mac OS X confuses the dynamic loader, because `fio_tls_missing.c` is built into the library defining the same symbols as `fio_tls_openssl.c` does, so in https://github.com/boazsegev/facil.io/commit/21614c0876769bbb83a4e46f65123e586171543e I renamed the file to `fio_tls_missing.sample-c` s.t. it is not built into the library anymore
- The other changes concern the `makefile`:
  - On OS X I installed OpenSSL-1.1 via [MacPorts](https://www.macports.org/), however facil.io's `makefile` didn't find it (it assumed standard empty compiler and linker flags whereas MacPorts installed into `/opt/local`). Thus, https://github.com/boazsegev/facil.io/commit/56fece97cefffe82561e17a2e43ab27bbe7ad726 uses `pkg-config` to determine the flags necessary to link against OpenSSL, which fixes that issue for me.
  - On Linux when building facil.io with GCC (v7.3.0, v8.3.0, v9.1.0, v9.2.0) and the default optimization `-O2` from the `makefile`, I encountered a strange bug where `http_upgrade2ws()` would silently return a negative value and not continue to upgrade the connection. I cannot be more precise at this point, would have to investigate. The issue does not occur with `clang-7` or `clang-8`. However, I tracked the bug to the `-fipa-icf` option of GCC (`-O1 -fipa-icf` already causes the buggy behaviour), therefore https://github.com/boazsegev/facil.io/commit/63745209efb7ea03fe18b7959a236e671e28c369 passes `-fno-ipa-icf` to the default `OPTIMIZATION` variable in the `makefile`.
  - https://github.com/boazsegev/facil.io/pull/71/commits/01edc0a3b697249b6eb6ca83af76e2e4c639ece4 is a small optimization of the `makefile` I made because (locally, not in this PR) I also wanted to build a static `libfacil.a` archive.
  - Linking against `libfacil.so` on OS X requires an `-install_name` to be set in the object file for OS X's dynamic loader to find the absolute path to the library when loading the program linking to it. https://github.com/boazsegev/facil.io/pull/71/commits/775e22202ccf5afb3d4807ce27303f7e8882719c adds the OS X specific linking options on Darwin.
  - Finally, since `pkg-config` worked quite well and the project I'm developing already relies on it for other libraries, https://github.com/boazsegev/facil.io/pull/71/commits/aa4c01c31ce0051814741f0ab30a5ecc6fb8758e makes the `makefile` learn how to write a `pkgconfig/facil.pc` file by itself, which makes linking to `libfacil.so` much easier for me.

I realize, you are developing v0.8 now, still I hope you find these patches useful. If you want to any changes made to them, please let me know.

### DIFF
--- a/lib/facil/tls/fio_tls_missing.c
+++ b/lib/facil/tls/fio_tls_missing.c
@@ -19,7 +19,7 @@ Feel free to copy, use and enjoy according to the license provided.
  */
 #include "fio_tls.h"
 
-#if 1 /* TODO: place library compiler flags here */
+#if !defined(HAVE_TLS) /* Library compiler flags */
 
 #define REQUIRE_LIBRARY()
 #define FIO_TLS_WEAK
--- a/lib/facil/tls/fio_tls_missing.c
+++ b/lib/facil/tls/fio_tls_missing.c
@@ -19,7 +19,7 @@ Feel free to copy, use and enjoy according to the license provided.
  */
 #include "fio_tls.h"
 
-#if !defined(HAVE_TLS) /* Library compiler flags */
+#if !defined(FIO_TLS_FOUND) /* Library compiler flags */
 
 #define REQUIRE_LIBRARY()
 #define FIO_TLS_WEAK
--- a/makefile
+++ b/makefile
@@ -524,8 +524,10 @@ build_objects: $(LIB_OBJS) $(MAIN_OBJS)
 
 lib: | create_tree lib_build
 
-lib_build: $(LIB_OBJS)
-	@$(CCL) -shared -o $(DEST)/libfacil.so $^ $(OPTIMIZATION) $(LINKER_FLAGS)
+$(DEST)/libfacil.so: $(LIB_OBJS)
+	@$(CCL) -shared -o $@ $^ $(OPTIMIZATION) $(LINKER_FLAGS)
+
+lib_build: $(DEST)/libfacil.so
 	@$(DOCUMENTATION)
 
 
--- a/makefile
+++ b/makefile
@@ -425,7 +425,7 @@ else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_BEARSSL_EXT), "-lbearssl"), 0)
 	LINKER_LIBS_EXT:=$(LINKER_LIBS_EXT) bearssl
 else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_OPENSSL), "-lcrypto" "-lssl"), 0)
   $(info * Detected the OpenSSL library, setting HAVE_OPENSSL)
-	FLAGS:=$(FLAGS) HAVE_OPENSSL
+	FLAGS:=$(FLAGS) HAVE_OPENSSL HAVE_TLS
 	LINKER_LIBS_EXT:=$(LINKER_LIBS_EXT) crypto ssl
 else
   $(info * No compatible SSL/TLS library detected.)
--- a/makefile
+++ b/makefile
@@ -88,6 +88,8 @@ ifndef CPPSTD
 	CPPSTD:=gnu++11
 endif
 
+PKG_CONFIG ?= pkg-config
+
 # for internal use - don't change
 LINKER_LIBS_EXT:=
 
@@ -414,6 +416,14 @@ int main(void) { \\n\
 # automatic library adjustments for possible BearSSL library
 LIB_PRIVATE_SUBFOLDERS:=$(LIB_PRIVATE_SUBFOLDERS) $(if $(wildcard lib/bearssl),bearssl)
 
+ifeq ($(shell $(PKG_CONFIG) -- openssl 2>&1 >/dev/null; echo $$?), 0)
+	OPENSSL_CFLAGS = $(shell $(PKG_CONFIG) --cflags openssl)
+	OPENSSL_LDFLAGS = $(shell $(PKG_CONFIG) --libs openssl)
+	OPENSSL_LIBS =
+endif
+
+OPENSSL_LDFLAGS ?= "-lssl" "-lcrypto"
+
 # add BearSSL/OpenSSL library flags (exclusive)
 ifdef FIO_NO_TLS
 else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_BEARSSL_SOURCE), $(EMPTY)), 0)
@@ -423,10 +433,12 @@ else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_BEARSSL_EXT), "-lbearssl"), 0)
   $(info * Detected the BearSSL library, setting HAVE_BEARSSL)
 	FLAGS:=$(FLAGS) HAVE_BEARSSL
 	LINKER_LIBS_EXT:=$(LINKER_LIBS_EXT) bearssl
-else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_OPENSSL), "-lcrypto" "-lssl"), 0)
+else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_OPENSSL), $(OPENSSL_CFLAGS) $(OPENSSL_LDFLAGS)), 0)
   $(info * Detected the OpenSSL library, setting HAVE_OPENSSL)
 	FLAGS:=$(FLAGS) HAVE_OPENSSL HAVE_TLS
-	LINKER_LIBS_EXT:=$(LINKER_LIBS_EXT) crypto ssl
+	LINKER_LIBS_EXT:=$(LINKER_LIBS_EXT) $(OPENSSL_LIBS)
+	LDFLAGS += $(OPENSSL_LDFLAGS)
+	CFLAGS += $(OPENSSL_CFLAGS)
 else
   $(info * No compatible SSL/TLS library detected.)
 endif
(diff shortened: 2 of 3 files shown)


boazsegev:
Hi @fbrausse ,

Thanks you for the PR!!!

The memory leak was a masterful catch, nice work!

To my shame, I already fixed the memory leak on the 0.8 branch and forgot to back port the fix (I got distracted with other projects).

However, I do have a few things I would like to clarify before I can merge the PR:

1. `pkg_config` isn't always available. For example, it's not installed by default on MacOS.

   I love the options you added and would love to keep them... but I don't want the `makefile` to **require** something that isn't there - but it can definitely test if it's there before using it.

   Maybe [the Makefile's TLS  `if` statement](https://github.com/boazsegev/facil.io/pull/71/files#diff-cd5dbf3629c558b0104ba8f0937d6816R433) could test for `pkg-config` && OpenSSL  in a different branch and keep the existing test for cases where `pkg-config` isn't available?

1. TLS and Weak Symbols:

    It's true that week symbols on dynamic libraries mean nothing, but they are helpful when facil.io is used as a static library or as a source-code library. This is especially true when OpenSSL isn't available or when facil.io is compiled without TLS support (i.e., when NginX takes care of the TLS layer).

    Any chance we could get the same result for the `fio_tls_missing` without renaming it?

   Maybe [using the `HAVE_OPENSSL` flag](https://github.com/boazsegev/facil.io/blob/b4d434aff3ee7059e75c0d242372264613de48fc/lib/facil/fio.c#L6394) instead of renaming the file?

   The behavior _should_ be that `fio_tls_missing` compiles (or applies) **only** when OpenSSL support is missing or removed.

Thank you again - cool work!
Bo.

fbrausse:
Hi Bo,

thank you for your thoughtful reply. I agree with your `pkg-config` reasoning and made the test conditional on whether `$(PKG_CONFIG) -- openssl` works. I've tried to keep the existing logic intact defining variables `OPENSSL_{CFLAGS|LDFLAGS|LIBS}' and using those. Please let me know whether that is close enough.

I've reverted the renaming and updated the patch series.

If I understand you correctly, you mean that the fio_tls_* symbols should always be defined, but just shouldn't be used when facil.io is *compiled without TLS support*. This is the case in 2 scenarios which you mention, right?
1. `FIO_NO_TLS` is not defined but no compatible TLS library has been found by the `makefile`: OK, I understand.
2. `$ make FIO_NO_TLS=1`: I'm confused here, isn't the intended meaning of this variable to *not* have TLS support?

As a conservative approach, the updated patch continues to ignore `FIO_NO_TLS` instead of e.g. setting a flag `WANT_TLS` and only then building `fio_tls_missing.c`. I've added a flag `HAVE_TLS` which is enabled only for OpenSSL at the moment, because it seems there are no other TLS backends available right now.

lib/facil/tls/fio_tls_missing.c
```diff
@@ -19,7 +19,7 @@ Feel free to copy, use and enjoy according to the license provided.
  */
 #include "fio_tls.h"
 
-#if 1 /* TODO: place library compiler flags here */
+#if !defined(HAVE_TLS) /* Library compiler flags */
```
boazsegev: The conservative approach was exactly right.

However - namespaces:

Please rename the flag to `FIO_TLS_FOUND`. This will mitigate possible name conflicts.

makefile
```diff
@@ -423,10 +440,14 @@ else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_BEARSSL_EXT), "-lbearssl"), 0)
   $(info * Detected the BearSSL library, setting HAVE_BEARSSL)
 	FLAGS:=$(FLAGS) HAVE_BEARSSL
 	LINKER_LIBS_EXT:=$(LINKER_LIBS_EXT) bearssl
-else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_OPENSSL), "-lcrypto" "-lssl"), 0)
+else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_OPENSSL), $(OPENSSL_CFLAGS) $(OPENSSL_LDFLAGS)), 0)
   $(info * Detected the OpenSSL library, setting HAVE_OPENSSL)
-	FLAGS:=$(FLAGS) HAVE_OPENSSL
-	LINKER_LIBS_EXT:=$(LINKER_LIBS_EXT) crypto ssl
+	FLAGS:=$(FLAGS) HAVE_OPENSSL HAVE_TLS
```
boazsegev: Same, please rename the `HAVE_TLS` flag to `FIO_TLS_FOUND`

makefile
```diff
@@ -423,10 +440,14 @@ else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_BEARSSL_EXT), "-lbearssl"), 0)
   $(info * Detected the BearSSL library, setting HAVE_BEARSSL)
 	FLAGS:=$(FLAGS) HAVE_BEARSSL
```
boazsegev: Please add a TODO comment, to add `FIO_TLS_FOUND` once BearSSL is supported (just so we remember what we did).

boazsegev:
> made the test conditional on whether `$(PKG_CONFIG) -- openssl` works.

Nicely done 👍🏻

> If I understand you correctly, you mean that the `fio_tls_*` symbols should always be defined, but just shouldn't be used when facil.io is compiled without TLS support.

Yes, exactly. The `fio_tls_*` symbols are crucial enough that in the 0.8.x release they will be part of the core library.

Attempting a TLS connection when `FIO_NO_TLS` was defined (or no library was found), should result in facil.io crashing the program with the message "No supported SSL/TLS library available" (but the program correctly compiles).

> `$ make FIO_NO_TLS=1`: I'm confused here, isn't the intended meaning of this variable to not have TLS support?

Yes 👍🏻

The idea is to force the compilation to behave as if no supported TLS library is available (even if it is).

The `fio_tls_*` symbols should still be defined, just as if no TLS library was found.

makefile
```diff
@@ -423,10 +430,13 @@ else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_BEARSSL_EXT), "-lbearssl"), 0)
...
 	LINKER_LIBS_EXT:=$(LINKER_LIBS_EXT) bearssl
-else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_OPENSSL), "-lcrypto" "-lssl"), 0)
+else ifeq ($(call TRY_COMPILE, $(FIO_TLS_TEST_OPENSSL), $(shell pkg-config --cflags --libs openssl)), 0)
```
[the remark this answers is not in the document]
fbrausse: Are you suggesting something like this?
```
   ...
else ifeq ([pkg-config check],0) && ([try-compile with pkg-config flags check])
   ...
else ifeq ([try-compile with fixed strings check])
   ...
```
In that case, for a user it would not be possible anymore to override these flags for `openssl` on the command line when the `[pkg-config check]` fails. I don't really need to, but it feels a bit inconsistent. (On a side note, I don't know how to do the `&&` in `GNU make`, sorry, I would break the structure of the `if`-`else`-chain.)

lib/facil/tls/fio_tls_missing.c
fbrausse: Thanks! There is no prefix because this flag is only used during build, not exported to users, I followed the surrounding style (`HAVE_OPENSSL`, `HAVE_BEARSSL`, etc.), but will push a new commit that renames this flag shortly.

makefile
fbrausse: Done

makefile
fbrausse: Done.

makefile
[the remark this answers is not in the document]
fbrausse: I am sorry, I was referring the flags `OPENSSL_CFLAGS`, `OPENSSL_LDFLAGS` in the new version.

lib/facil/tls/fio_tls_missing.c
boazsegev: Perfect, thanks for the quick fix :)

FYI: the `HAVE_*` style is the `configure` utility style. I kept it only for library/feature names that could be share across libraries. It's unlikely, but it might conflict with other flags when multiple libraries are used (in the `./lib` folder).

makefile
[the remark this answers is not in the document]
boazsegev: Oh, no, this comment was here from before you fixed it. I loved your fix, more readable than my original suggestion.

boazsegev:
@fbrausse Thanks!!!
🎉👏🏻👏🏻👏🏻🎉
👍🏻👍🏻👍🏻👍🏻👍🏻

fbrausse:
Great! Thank you very much :)
I'm looking forward to v0.8!

PR #76, opened 2019-10-03
Bugfix compiling C++

joeyhoek:
I came across this issue whilst creating a router for an API in my most recent project.
For me it was easily fixed by adding types.

Fixed issue compiling cxx14
Example: error: non-constant-expression cannot be narrowed from type
'off_t' (aka 'long') to 'uintptr_t' (aka 'unsigned long') in initializer
list [-Wc++11-narrowing]

### DIFF
--- a/lib/facil/fio.h
+++ b/lib/facil/fio.h
@@ -1250,7 +1250,7 @@ inline FIO_FUNC ssize_t fio_write(const intptr_t uuid, const void *buffer,
 inline FIO_FUNC ssize_t fio_sendfile(intptr_t uuid, intptr_t source_fd,
                                      off_t offset, size_t length) {
   return fio_write2(uuid, .data.fd = source_fd, .length = length, .is_fd = 1,
-                    .offset = (unsigned long)offset);
+                    .offset = (uintptr_t)offset);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


boazsegev:
Hi @joeyhoek ,

Thank you for the PR 🙏🏻👍🏻

I don't often use C++, so I'm very thankful for the input.

However, I can't help wondering if the types you chose aren't system specific...

The `offset` type in `fio_write2` (`uintptr_t`) might not map to `unsigned long` on every system.

Also, according to the specifications, `tv_sec` is typed as `time_t`, whereas you casted it to `long`. I assume it would be better to switch that to `time_t`, so it's more portable.

(Note: `tv_nsec` should be typed as `long` by the specification, so no worries about that one)

Anyway, could you fix those before I merge?

Thanks!
    Bo

joeyhoek:
Hi @boazsegev,

I think you meant this?

Love the work btw. I'm not very comfortable with C per se, so unfortunately don't think I can be a big help there. If you need any help in the webdevelopment part etc, I'd be happy to help out.

boazsegev:
Hi @joeyhoek ,

Yes, perfect 👍🏻 Thanks🙏🏻

Thank you for helping improve C++ compatibility ☺️

B.

PR #85, opened 2020-01-20
fix deprecated API usage in boiler_plate example

utoni:
- use fio_url_s type instead of deprecated http_url_s

Signed-off-by: Toni Uhlig <matzeton@googlemail.com>

### DIFF
--- a/examples/boiler_plate/src/cli.c
+++ b/examples/boiler_plate/src/cli.c
@@ -73,7 +73,7 @@ void initialize_cli(int argc, char const *argv[]) {
   if (fio_cli_get("-redis") && strlen(fio_cli_get("-redis"))) {
     FIO_LOG_INFO("* Initializing Redis connection to %s\n",
                  fio_cli_get("-redis"));
-    http_url_s info =
+    fio_url_s info =
         http_url_parse(fio_cli_get("-redis"), strlen(fio_cli_get("-redis")));
     fio_pubsub_engine_s *e =
         redis_engine_create(.address = info.host, .port = info.port,
--- a/lib/facil/http/http.h
+++ b/lib/facil/http/http.h
@@ -988,7 +988,11 @@ typedef fio_url_s http_url_s
  *
  * Invalid formats might produce unexpected results. No error testing performed.
  */
-#define http_url_parse(url, len) fio_url_parse((url), (len))
+static inline fio_url_s __attribute__((deprecated("use fio_url_parse instead")))
+http_url_parse(const char *url, size_t length)
+{
+    return fio_url_parse(url, length);
+}
 
 #if DEBUG
 void http_tests(void);
--- a/lib/facil/http/parsers/http1_parser.c
+++ b/lib/facil/http/parsers/http1_parser.c
@@ -583,13 +583,13 @@ re_eval:
                                               : args->on_request)(args->parser))
       goto error;
     args->parser->state =
-        (struct http1_parser_protected_read_only_state_s){0, 0, 0};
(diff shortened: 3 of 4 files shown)


boazsegev:
Hi @lnslbrty ,

Thank you for your PR 🙏🏻

Nice catch on the deprecation warning.

However, I think we will need to update the function call as well as the type (using [`fio_url_parse`](https://github.com/boazsegev/facil.io/blob/c1306722fd84a23447671619fc76eb8a11a6cb39/lib/facil/fio.h#L866-L915) instead of `http_url_parse`).

Could you fix that before I merge?

examples/boiler_plate/src/cli.c
```diff
@@ -73,7 +73,7 @@ void initialize_cli(int argc, char const *argv[]) {
   if (fio_cli_get("-redis") && strlen(fio_cli_get("-redis"))) {
     FIO_LOG_INFO("* Initializing Redis connection to %s\n",
                  fio_cli_get("-redis"));
-    http_url_s info =
+    fio_url_s info =
```
boazsegev: I think we will need to update the function call as well as the type (using [`fio_url_parse`](https://github.com/boazsegev/facil.io/blob/c1306722fd84a23447671619fc76eb8a11a6cb39/lib/facil/fio.h#L866-L915) instead of `http_url_parse`).

lnslbrty:
Will do that ASAP.

boazsegev:
Cool 😎 

I'm mostly offline for the next week or two, but I'll try to merge as soon as I can :-)

lnslbrty:
Fixed.
I would also recommend to remove the broken  **makefile** since all (modern) *Nix OSes support CMake (which is IMHO state-of-the-art technology).
I also need to investigate an issue regarding some explicit function pointer casts before I start porting *facil.io* to *OpenWrt* (and ofc maintaining the pkg).

boazsegev:
Hi @lnslbrty ,

Thank you for the PR and the updates 👍🏻

> Fixed.

I'll probably merge it soon. My online time is still limited right now.

> I would also recommend to remove the broken makefile since all (modern) *Nix OSes support CMake (which is IMHO state-of-the-art technology).

As for the CMake - I don't know why people use it. I find GNU `make` much easier to manage. However, I noticed you claim that the `makefile` is broken... how so?

> I also need to investigate an issue regarding some explicit function pointer casts before I start porting facil.io to OpenWrt (and ofc maintaining the pkg).

Yes... facil.io uses some virtual function tables and pointer casting that walks a thin line by assuming that all pointers are an equal type-size (`sizeof(uintptr_t)`) - which might not be true on some older embedded systems (where function pointers are one length and type pointers are another length). 

If you have any specific concerns I can help with, please let me know.

Kindly,
  Bo.

lnslbrty:
> As for the CMake - I don't know why people use it. I find GNU `make` much easier to manage. However, I noticed you claim that the `makefile` is broken... how so?

A simple `make` results in a missing main function error. I do not see any compile/linker commands since they are all "silenced". I would not recommend to do such things in Makefiles as it takes a lot more time to debug errors.
Another thing is the not-easy-to-read variable substitution. On my machine `MAINSRC`/`MAIN_OBJS` gets always expanded to an empty string.

> Yes... facil.io uses some virtual function tables and pointer casting that walks a thin line by assuming that all pointers are an equal type-size (`sizeof(uintptr_t)`) - which might not be true on some older embedded systems (where function pointers are one length and type pointers are another length).

I'm working on it (but currently OOT - OutOfTime Exception). :smile:

PR #86, opened 2020-01-25
add shared library support

lnslbrty:
Signed-off-by: Toni Uhlig <matzeton@googlemail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,9 @@
-project(facil.io C)
+project(facilio C)
 cmake_minimum_required(VERSION 2.4)
 
 find_package(Threads REQUIRED)
 
-set(facil.io_SOURCES
+set(facilio_SOURCES
   lib/facil/fio.c
   lib/facil/tls/fio_tls_missing.c
   lib/facil/tls/fio_tls_openssl.c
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
-project(facil.io C)
+project(facilio C)
```
boazsegev: Is there a reason to drop the `.`?

If so, I'd like to know.

CMakeLists.txt
```diff
@@ -42,3 +42,6 @@ target_include_directories(facil.io
   PUBLIC  lib/facil/redis
 )
 
+set_property(TARGET facilio_objlib PROPERTY POSITION_INDEPENDENT_CODE 1)
```
boazsegev: Interesting... I'm no CMake user / expert - what do these lines do?

Does this replace the option of using facil.io as a source code library?

CMakeLists.txt
lnslbrty: It is confusing (and probably not best-practice) using a `.` for library names.
This character is usually reserved as separator between library name, library suffix and version.

CMakeLists.txt
lnslbrty: It basically adds a `CFLAG` named `-fPIC` which tells the compiler to generate position independent code.
For shared libraries this is a must as the start address of the memory mapped library can be different to the default image base address set by the ELF linker.

lnslbrty:
> I assume that adding shared / static library support does not remove the ability to use facil.io as a code library?

It won't. There is still a static library named `libfacilio_static.a` (using `*_static.a` is a common design pattern for CMake based projects).

PR #92, opened 2020-06-21
Fix theme toggle href function

GnaneshKunal:
- Prevent redirection by not returning anything from the theme
  function.

Before:
output

After:
output

From: https://stackoverflow.com/a/40048197/6830396
> "...the href of a link is that normally, a javascript: URL will redirect the browser to a plain text version of the result of evaluating that JavaScript. ..."


boazsegev:
@GnaneshKunal ,

Thank you for the PR 🎉👏🏻🙏🏻👍🏻

I'm not experiencing this issue on Brave or Safari (`return false` seems to negate the reload)... however, it's obvious that the issue occurs on the browser you're using (though I don't recognize it).

Thanks again!
   Bo.

boazsegev:
FYI:

I'm not sure which browser you're using, but a refresh shouldn't be happening any case, since the `onclick` function shouldn't effect an `href` attribute.

Note that the resulting HTML code is:

```html
<a href="#" id="theme" onclick="change_themes()">Day Theme</a>
```

And the JavaScript function has an:

```js
function change_themes(e) {
   // ...
   if(e) e.preventDefault();
};
```

In theory, not only should the `onclick` stop any further processing, but once the `onclick` callback is finished (assuming processing continues), the `href` attribute would determine the action (here, it is directing the page to an anchor rather than performing a page refresh).

I would offer the opinion that the browser you're using doesn't follow the standards as well as it should.

B.

GnaneshKunal:
@boazsegev 

Thank you for the merge.

> I'm not sure which browser you're using

I was using firefox.

This is from the standards: : https://stackoverflow.com/a/40048197/6830396

> "...the href of a link is that normally, a javascript: URL will redirect the browser to a plain text version of the result of evaluating that JavaScript. ..."

> > Note that the resulting HTML code is:
```js
<a href="#" id="theme" onclick="change_themes()">Day Theme</a>
```
> > And the JavaScript function has an:
```js
function change_themes(e) {
   // ...
   if(e) e.preventDefault();
};
```

That is right. But that is not exactly what was in the docs code. Instead of embedding the function on `onclick` we have embeded it directly in the `href`.

From your above example, you have put `#` in the href and `event.preventDefault()` to stop the event propagation but in the docs code it was 
```html
<a href="javascript:change_themes()">Change theme</a>
```

If you notice here, **there's no onlick (but href function)** and the result of the href function is where the page will be redirected do. If change themes returns undefined it won't redirect but if it returns some other value, then the page will be redirected to the string representation of that value. We return `false` in the `change_themes` function and it doesn't stop the page from redirection but the value is treated as a link and the page is redirected to the string representation of the value.

boazsegev:
@GnaneshKunal ,

Thank you for your update. I'm keeping it because it obviously works around an issue on some browsers... however - just for your reference:

According to [StackOverflow, the best option it you use](https://stackoverflow.com/questions/1070760/javascript-function-in-href-vs-onclick#11348403):

```html
<a id="link_id" title="Foo"  href="no_javascript.html" onclick="foo();return false;">foo</a>
```

This approach is commonly regarded as best. See also the [User Experience forum](https://ux.stackexchange.com/a/3920).

In addition, FireFox on macOS works as expected, without reloading the page (maybe you need to update your browser?).

Anyway, I'll keep your update and I'm thankful that you exposed the need for a workaround.

Kindly,
   Bo.

PR #96, opened 2020-11-01
Fix for LP32 platforms: convert pointers via uintptr_t to/from integral types

fbrausse:
Hi,

I'm using facil.io on a i686 Linux machine (gcc-9.3.0) and noticed lots of warnings about casting pointers to integers of different widths, e.g. `(uint64_t)&obj`, an (undefined) left-shift of a 32-bit `size_t` by 33 in the risky hash computation, as well as the explicit use of `sendfile64` with an `off_t` parameter.

This patch fixes these warnings by
* casting pointers to `uintptr_t` first,
* using an intermediate `uint64_t` for the risky hash mixing in of `size_t len`, and
* unconditionally adding the `_FILE_OFFSET_BITS=64` #define to the `FLAGS` in `makefile`.

The API has not been touched, just the internal calls. This has been the most obvious way for me to fix compilation of facil.io on this 32-bit platform, though I'm open to fixing these in a different way if you have any suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Change Log
 
+**Fix**: fixed some 32 bit compatibility concerns. Credit to Franz Brausse ( @fbrausse ) for the [PR @ boazsegev/facil.io#96](https://github.com/boazsegev/facil.io/pull/96).
+
 ### v. 0.7.5 (2020-05-18)
 
 **Security**: backport the 0.8.x HTTP/1.1 parser and it's security updates to the 0.7.x version branch. This fixes a request smuggling attack vector and Transfer Encoding attack vector that were exposed by Sam Sanoop from [the Snyk Security team (snyk.io)](https://snyk.io). The parser was updated to deal with these potential issues.
--- a/lib/facil/cli/fio_cli.c
+++ b/lib/facil/cli/fio_cli.c
@@ -50,7 +50,7 @@ typedef struct {
 #define AVOID_MACRO
 
 #define FIO_CLI_HASH_VAL(s)                                                    \
-  fio_risky_hash((s).data, (s).len, (uint64_t)fio_cli_start)
+  fio_risky_hash((s).data, (s).len, (uintptr_t)fio_cli_start)
 
 /* *****************************************************************************
 CLI Parsing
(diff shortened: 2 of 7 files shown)


boazsegev:
Hi @fbrausse ,

Thanks for the PR! **Could you add a note in the CHANGELOG and credit yourself for the PR?** ([like the one added here](https://github.com/facil-io/cstl/commit/f03f914803fed5d4d3e5b8a9ac9a98ffbb156616)).

I already implemented some of the proposed changes in [one of the 0.8.0.beta repositories](https://github.com/facil-io/cstl) (this is just a single header library with the core modules that could be isolated from the IO logic)... I should have thought to test the code on 32 bit platforms, but all the machines I run my code on are 64 bit 🤷‍♂️

Anyway, thanks a lot!

Bo.

P.S.

**If there's anything you want to see in the next version (where breaking changes to the API will occur), let me know.**

I hate breaking the API, but 0.7.x is such a patchwork of ideas and the API isn't as unified and coherent as I'd like. I'm trying to smooth out API inconsistencies - i.e., [`fio_str_free`](http://facil.io/0.7.x/fio#fio_str_free) should really have matched [`fiobj_free`'s  behavior](http://facil.io/0.7.x/fiobj_core#fiobj_free) rather than having it behave the same as [`fiobj_str_resize(str, 0)`](http://facil.io/0.7.x/fiobj_str#fiobj_str_resize)... so this is a good time to voice out any comments / ideas.

fbrausse:
Hi Bo,

I've added the Changelog entry as requested and expanded detail in the commit message, the rest of the patch remains unchanged.

Thank you for all your work on these libraries and APIs in particular - we don't need Python to have simple interfaces! :)

Regarding API consistency, I'm afraid there are no suggestions I can give you right now. However, for ref-counted types (like `fio_str`) I agree that `*_free()` should behave like the counter-part to `*_dup()` as in C (nearly) all we have are naming conventions. 

Personally, I've adopted the convention of `ref()`/`unref()` pair for ref-counted types (from GObject, I believe) vs. the `init()`/`fini()` pair for types with directly managed lifetimes to be created on the stack. My heap-allocated types (by `create()`/`destroy()` for alloc + init) usually are not ref-counted. If they are, I use `create()`/`ref()`/`unref()` without `destroy()`.

Kind regards,
Franz

boazsegev:
Great! thanks!

PR #102, opened 2021-06-16
Make tests run by default

janbiedermann:
Fixing "cast between incompatible function types from" for good would need some changes.
For example:
 changing
```
 struct fio_packet_s {
  fio_packet_s *next;
  int (*write_func)(int fd, struct fio_packet_s *packet);
  void (*dealloc)(void *buffer);
  union {
    void *buffer;
    intptr_t fd;
  } data;
  uintptr_t offset;
  uintptr_t length;
};
```

to:
```
 struct fio_packet_s {
  fio_packet_s *next;
  int (*write_func)(int fd, struct fio_packet_s *packet);
  union {
    void (*dealloc_buf)(void *buffer);
    void (*dealloc_fd)(intptr_t fd);
  } dealloc;
  union {
    void *buffer;
    intptr_t fd;
  } data;
  uintptr_t offset;
  uintptr_t length;
};
```
and match references accordingly. I would implement that, if that is ok for you?

### DIFF
--- a/lib/facil/fio.c
+++ b/lib/facil/fio.c
@@ -3917,23 +3917,31 @@ void fio_start FIO_IGNORE_MACRO(struct fio_start_args args) {
   fio_data->is_worker = 0;
 
   fio_state_callback_force(FIO_CALL_PRE_START);
+#if HAVE_OPENSSL
   FIO_LOG_INFO(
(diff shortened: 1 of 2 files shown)


janbiedermann:
ill recreate this one

PR #106, opened 2021-07-06
Windows support for 0.7.x

janbiedermann:
see #104 for discussion so far

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,9 +11,12 @@ lib/bearssl/
 .gitignore
 .DS_Store
 .gitmodules
+.vscode
 
 callgrind*
 docs/assets/styles/main.css.map
 tests/xxhash.h
 tests/xxhash.c
 tests/words.txt
+obj.obj
+a.exe
--- a/README.md
+++ b/README.md
@@ -24,6 +24,17 @@ The master branch on the `git` repo is the development branch and is likely to b
 
 Please select a release version for any production needs.
 
+### Windows support
+
+Windows support is available for development and not at all recommended for production use.
+Only one worker is allowed, because there is no fork() support on Windows.
+Instead of unix sockets a local tcp port in the port range from 10000 to 19999 is used for inter process/thread communication, which makes things a bit slower.
+
+To get started:
+1. Install Ruby (in general a great idea) with the Devkit from [rubyinstaller.org](https://rubyinstaller.org/downloads/) and make sure, you have the commands in the path. (Or alternatively install MingW64 for Windows.)
+2. Open a cmd and enable MingW64 from the Ruby Devkit with: `ridk enable`. (Or use the appropriate command from MingW64.)
+Ready to go ...
+
 ### Who's running on `facil.io`
 
 * [Iodine, a Ruby HTTP/Websockets Ruby application server](https://github.com/boazsegev/iodine) is powered by `facil.io` - so everyone using the iodine server is running on facil.io.
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Please select a release version for any production needs.
 ### Windows support
 
 Windows support is available for development and not at all recommended for production use.
-There si no ssl support.
+There is no ssl support.
 Only one worker is allowed, because there is no fork() support on Windows.
 Instead of unix sockets a local tcp port in the port range from 10000 to 19999 is used for inter process/thread communication, which makes things a bit slower.
 
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Please select a release version for any production needs.
 ### Windows support
 
 Windows support is available for development and not at all recommended for production use.
+There si no ssl support.
 Only one worker is allowed, because there is no fork() support on Windows.
 Instead of unix sockets a local tcp port in the port range from 10000 to 19999 is used for inter process/thread communication, which makes things a bit slower.
 
--- a/examples/benchmarks/k6bench.js
+++ b/examples/benchmarks/k6bench.js
@@ -1,0 +1,37 @@
+import ws from "k6/ws"; 
+
+export default function() { 
+
+  let url = "ws://localhost:3000";  
+  let params = { "tags": { "my_tag": "hello" } };   
+
+  let res = ws.connect(url, params, function(socket) {  
+
+      socket.on('open', function open() {   
+          console.log('connected'); 
+
+          socket.setInterval(function timeout() {   
+            socket.ping();  
+            console.log("Pinging every 1sec (setInterval test)");   
+          }, 1000); 
+      });   
+
+      socket.on('ping', function () {   
+          console.log("PING!"); 
+      });   
+
+  
+      socket.on('pong', function () {   
+          console.log("PONG!"); 
+      });   
+
+      socket.on('close', function() {   
+          console.log('disconnected');  
+      });   
+
+      socket.setTimeout(function () {   
+          console.log('2 seconds passed, closing the socket');  
+          socket.close();   
+      }, 2000); 
+});
+}
--- a/lib/facil/fio.c
+++ b/lib/facil/fio.c
@@ -4,6 +4,10 @@ License: MIT
 
 Feel free to copy, use and enjoy according to the license provided.
 ***************************************************************************** */
+#ifdef __MINGW32__
+#define FIO_FORCE_MALLOC 1
+#define FIO_DISABLE_HOT_RESTART 1
+#endif
 
 #include <fio.h>
 
@@ -18,9 +22,12 @@ Feel free to copy, use and enjoy according to the license provided.
 #include <errno.h>
 #include <limits.h>
 #include <pthread.h>
+#ifndef __MINGW32__
 #include <sys/mman.h>
+#endif
 #include <unistd.h>
 
+#ifndef __MINGW32__
 #include <netdb.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
@@ -29,12 +36,27 @@ Feel free to copy, use and enjoy according to the license provided.
 #include <sys/ioctl.h>
 #include <sys/resource.h>
 #include <sys/socket.h>
+#endif
 #include <sys/stat.h>
 #include <sys/types.h>
+#ifndef __MINGW32__
 #include <sys/un.h>
 #include <sys/wait.h>
 
 #include <arpa/inet.h>
+#endif
+
+#ifdef __MINGW32__
+#include <windef.h>
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#endif
+
+#if HAVE_OPENSSL
+#include <openssl/bio.h>
+#include <openssl/err.h>
+#include <openssl/ssl.h>
+#endif
 
 #if HAVE_OPENSSL
 #include <openssl/bio.h>
@@ -47,12 +69,14 @@ Feel free to copy, use and enjoy according to the license provided.
 #define FIO_ENGINE_POLL 0
 #endif
 
-#if !FIO_ENGINE_POLL && !FIO_ENGINE_EPOLL && !FIO_ENGINE_KQUEUE
+#if !FIO_ENGINE_POLL && !FIO_ENGINE_EPOLL && !FIO_ENGINE_KQUEUE && !FIO_ENGINE_WSAPOLL
 #if defined(__linux__)
 #define FIO_ENGINE_EPOLL 1
 #elif defined(__APPLE__) || defined(__FreeBSD__) || defined(__NetBSD__) ||     \
     defined(__OpenBSD__) || defined(__bsdi__) || defined(__DragonFly__)
 #define FIO_ENGINE_KQUEUE 1
+#elif defined(__MINGW32__)
+#define FIO_ENGINE_WSAPOLL 1
 #else
 #define FIO_ENGINE_POLL 1
 #endif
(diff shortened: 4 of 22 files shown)


janbiedermann:
I believe -mthreads must be used, to get multithreading compatible headers and libs on windows.

janbiedermann:
To install msys/mingw without ruby: https://www.msys2.org/

janbiedermann:
Working.
Remaining issues to be fixed later on:
Websocket connection drops, also on Linux with poll() and OS malloc
possible crashes for unknown reasons, also on Linux with poll() and OS malloc

Not perfect, has issues, but at the same time i fixed several crash bugs on the way, so overall improved

I think ill take that and port it to iodine and will see how it behaves.

I would like to suggest macros:
FIO_ASSERT_SOCKET
FIO_CLOSE_SOCKET
These can be used at a couple of places, saving some #ifdefs

Maybe ill implement those in the next days.

janbiedermann:
For thread suspend/wake up i use condition variables on Windows. This is silly, but it kinda works. Reason for this is, that the existing pipe and poll impementation for signalling somehow does not work reliably on Windows, could be because the pipe implementation of MingW is not up to spec.

boazsegev:
I'm still learning and can't reliably review the code yet (very frustrating for me not to be able to understand some of the Windows related details)... however:

## `FIO_ASSERT_SOCKET` / `FIO_CLOSE_SOCKET`

Yes, definitely. I took this approach myself while trying to incorporate you suggestions to the upcoming 0.8.x code base.

In addition, I found it also useful to wrap the system calls for read and write (since windows in unfriendly with them)... You can have a look at the [facil.io CSTL (C Simple Types Library)](https://github.com/facil-io/cstl) and try the HTTP test code ([`make test/http`](https://github.com/facil-io/cstl/blob/master/tests/http.c)).

## Windows SOCKET (`fd`) values (`int`?)

It seems to me, from testing different function calls against the windows kernel, that the Windows kernel keeps all `HANDLE`s (including `SOCKET` in an internal Array. Unlike POSIX, this internal Array seems to be global rather than process specific and includes all sorts of types of handles...

...however, this also puts us at a wonderful position where we can (possibly) assume that the value of these handles will not exceed a 31 bit value (a positive `int`). This might allow us to both shave a few bytes per connection (use `int` instead of `SOCKET`) as well as use direct access for the `fd_data` array.

We will need to test a `SOCKET` value against `caps` and refuse socket handles with values that exceed our capacity (which is now more of a map style capacity than a connection capacity), but it would keep the O(1) performance we have for the `fd_data` array.

## WSYS2 and `pread`

I'm not sure how you solved the `pread` part. Adding a function declaration didn't really help on my Windows 10 VM. Maybe I'm doing something wrong?

I found myself having to add a wrapper function that emulates `pread` ([example here](https://github.com/facil-io/cstl/blob/78a17beb575fbee44b9329f887b55658bd16dbf4/stl_slices/000%20header.h#L680-L718)) and honestly I have strong doubts if this is what I need to do to run your branch...

janbiedermann:
pread is already there https://github.com/janbiedermann/facil.io/blob/e22b099017a8c97d53ecc7d34eea2d595115a697/lib/facil/fio.c#L246

janbiedermann:
Windows SOCKET (fd) values (int?)
I recommend to use SOCKET and INVALID_SOCKET for testing for a valid socket. Testing for -1 with a int missed invalid sockets, thats why did it the other way. This may be a compiler/byte order/casting/whatever issue, i dont know, using SOCKET and testing for INVALID_SOCKET just works.

janbiedermann:
> I'm still learning and can't reliably review the code yet (very frustrating for me not to be able to understand some of the Windows related details)... 

Yeah, me too. For your allocator i found something:  https://docs.microsoft.com/en-us/windows/win32/memory/allocating-virtual-memory

Maybe thats helpful, because it operates on pages.

janbiedermann:
Actually i tried to avoid the Winapi as much as possible, to keep the footprint low and make maintenance easier.
Anyway, for thread handling there is this handy function available `pthread_getw32threadhandle_np()` , gets a Windows thread handle from a pthreads thread. This may be useful for suspend/resume/signalling

boazsegev:
### SOCKET and INVALID_SOCKET

I agree that we need to test using these tools, otherwise we might hit snags. But we should also test if `SOCKET <= 0x3FFFFFFFULL` If it is, we can safely store the value in an `int` and keep the code base the same for POSIX, including the `uuid` protection layer which adds a counter to the `fd` in order to prevent long running tasks from sending data to a recycled `fd`.

janbiedermann:
SOCKET and INVALID_SOCKET

well, you just can hope that nobody will run facil.io on a Windows Server/Datacenter Edition:
https://docs.microsoft.com/en-us/windows-server/get-started-19/editions-comparison-19
and
https://social.msdn.microsoft.com/Forums/en-US/dca7e775-b6c8-4b24-b75b-5799d8091836/does-server-20032008-have-tcp-connection-limits-like-xp-and-vista

"It depends on the edition, Web and Foundation editions have connection limits while Standard, Enterprise, and Datacenter do not."

note the "unlimited" and "no limit" ;-)

On my route to world domination i surely would like to have that as an option, just in case ;-D

boazsegev:
@janbiedermann ,

#### SOCKET and INVALID_SOCKET (thinking out loud, but in the end you're probably right):

```c
/* Considering limiting the SOCKET to 4 byte limits of int...? */
#ifndef FIO_SOCK_FD_ISVALID
#define FIO_SOCK_FD_ISVALID(fd) ((size_t)fd <= (size_t)0x7FFFFFFF)
#endif
```
You are absolutely right about this consideration, I also feel slightly guilty about relaying on an assumed internal implementation detail in the Windows kernel.

On one hand, if we conform to an opaque `SOCKET` type (a 64 bit value that can use all bits) than we are safely compatible with windows (which currently run on [about 25% of the public servers on the internet](https://en.wikipedia.org/wiki/Usage_share_of_operating_systems#Public_servers_on_the_Internet), with about 75% being POSIX machines)... that's important.

The price, of course, would be direct access to the `fd_data` array, the possible use of a hash map to link between the two data sets and more code changes. Maybe this is a fair trade, or maybe the performance will tank (and people will choose better alternatives).

On the other hand, 0x7FFFFFFF (2,147,483,647) open file descriptors are likely to be (currently) pushing any machine to it's physical limits... not to mention that it would be creating a single point of failure on any web architecture (I don't think any architecture designer would choose that)... so I doubt if the term "unlimited" really matters.

The question is more about relying on an internal implementation detail that can be observed but not validated and might change in the future 🤔

... in this sense I know that you are absolutely right and this is a **big** corner to cut... still, how likely is to happen (an `fd` that high)? and how do we deal with an endless (unlimited) amount of connections? 🤷🏼‍♂️

janbiedermann:
Yeah, alright. As i am mainly interested  to get iodine running for development. I think it is be safe to assume that the Windows desktop will always be limited, so that MS can sell server licenses.

boazsegev:
Thanks for the tip about the [The VirtualAlloc function](https://docs.microsoft.com/en-us/windows/win32/memory/allocating-virtual-memory). I incorporated your ideas and some of your code into the [facil.io CSTL (C Simple Types Library)](https://github.com/facil-io/cstl) which now supports the native Windows compiler (MSVC) on 64 bit systems (when compiling for 64 bit).

I tweaked your ideas for [`kill`, `pread` and `pwrite`](https://github.com/facil-io/cstl/blob/20b7860468ac814c6343444be0ed892c2dd55ce5/stl_slices/001%20patches.h#L50-L330) and used some others suggestions I picked up from your code. it seems to work well.

You can view the new memory allocator perform with your suggested Windows support at the [CSTL repo](https://github.com/facil-io/cstl) running:

```dos
cls && cl /Ox  /I. tests\ malloc.exe.c && malloc.exe 
``` 

The HTTP tests works on my VM as well:

```dos
cls && cl /Ox  /I. tests\ malloc.exe.c && malloc.exe 
``` 

Now I'm trying to work out a design that avoids the `fd_data` array so we can support Windows HANDLE sockets with values up to  31 bits (0x7FFFFFFF or 2,147,483,647). Hopefully it will produce decent performance.

I'm not going 64 bit socket HANDLEs just yet, but we might get there in time.

Cheers and thanks for the code. I'm still working on my end to make this work.

janbiedermann:
Fantastic! I will try your code. Well, thank you for your work! I like facil very much! 
🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻🙏🏻

boazsegev:
Thanks!

Please note that you'll be reviewing the lower level building blocks for the newly designed IO core (where everything will either work together or fall apart) ...

... the facil.io CSTL is just the building blocks for the house of cards that is the upcoming new IO core. It has a bunch of separate modules, such as thread abstractions, atomics, hash maps, dynamic array types, etc'... but there's no reactor, no event loop - just the parts that make them.

janbiedermann:
It seems the race condition that is causing crashes in fio_flush, is between fio_force_close and fio_flush. Above commit doesnt solve this race condition, but it properly provides errnos for the code depending on it, so that fio_force_close is not called from read. I tried to fix the race condition with some locks, but have not been successful this way. Anyway, so far no more crashes on Windows.

You wan to pick the errnos for recv for the building blocks for the house of cards.

boazsegev:
> It seems the race condition that is causing crashes in fio_flush, is between fio_force_close and fio_flush

I think the multi threaded design is weighing heavily on the system.

The race conditions don't only carry risks of code bugs and add performance concerns where there is high contention, but also allow user code to block facil.io services such as responding to `on_ready` events, timeouts (i.e. sending `pings`), etc'.

The new house of cards will probably have only one thread handling all "system" operations and "user" threads for user code. API calls will mostly schedule events on the system threads without actually doing anything (unless it is safe to do something atomically, perhaps something like changing the protocol).

janbiedermann:
For use within iodine:
https://spin.atomicobject.com/2015/09/28/windows-sockets-ruby-extensions/

Ruby is doing its own socket <-> handle mapping, which makes WSAPoll() fail, because it gets the ruby fd instead of the handle.
Not sure what to do.

boazsegev:
> Ruby is doing its own socket <-> handle mapping, which makes WSAPoll() fail, because it gets the ruby fd instead of the handle.

There's some interesting code to learn from at [their `win32.h` header file](https://github.com/ruby/ruby/blob/d92f09a5eea009fa28cd046e9d0eb698e3d94c5c/include/ruby/win32.h).

The interesting part, IMHO, is that they are basically using macros to wrap the original symbols rather than overwrite the function names. This seems to have been [true also before Ruby 2.0](https://github.com/ruby/ruby/blob/ruby_1_9_2/include/ruby/win32.h).

I think this means that if we avoid having the Ruby header mess with the facil.io library (include Ruby after facil.io) that we should be fine.

The [ENet library in the blog post](https://github.com/Dahrkael/rENet/tree/master/ext/renet/enet) a[[ears to be a header library. I imagine it was harder to isolate it from the preprocessor macros, not to mention that the author [included the library after including Ruby](https://github.com/Dahrkael/rENet/blob/7ece24f1b41603dcca23ef97cdb9dc6b1d5917cd/ext/renet/renet.h#L22-L23), allowing the Ruby macros to effect the ENet library.

Anyway, I think we'll be okay. We will need to be watchful, but with correct include file ordering and perhaps a few patches to the hijacking logic we should be fine... I hope.

janbiedermann:
> I think this means that if we avoid having the Ruby header mess with the facil.io library (include Ruby after including facil.io) then we should be fine.

That doesnt seem to work or my attempts were faulty. I went with getting the function pointers from the dll as suggested in the article, that works. But a new problem occured, a crash when accessing 
```
static __thread FIOBJ each_at_key = FIOBJ_INVALID;
```
in fiobj_hash.c
There may be other, similar cases. What to do?
I tried several compiler/linger options, no success.

janbiedermann:
Alright, i found the problem. When using gcc TLS via __thread from within a ruby thread on Windows it wont work, as Ruby uses TLS via the separate Windows functions https://github.com/ruby/ruby/blob/550b02e4790c406450008e3bbbf28d8982cc0908/thread_pthread.h#L80 . So when running within Ruby threads, facil.io must do the same, when creating its own threads via pthreads gcc __thread will work as usual.

janbiedermann:
rather here https://github.com/ruby/ruby/blob/550b02e4790c406450008e3bbbf28d8982cc0908/thread_win32.c#L135

boazsegev:
> When using gcc TLS via __thread from within a ruby thread on Windows it wont work

I thought that was a [pre-Windows-Vista concern. Even though facil.io will probably be loaded using `LoadLibrary`](https://docs.microsoft.com/en-us/cpp/parallel/thread-local-storage-tls?view=msvc-160&viewFallbackFrom=vs-2019) (for iodine), I hoped Windows 10 would work with [a native wrapping for `__thread`](https://github.com/facil-io/cstl/blob/48351b7b7beedd4ed67f02fa7ca3f5a3dfb0cd26/stl_slices/000%20header.h#L186-L190).

Using explicit thread storage would add code complexity that I hoped to avoid. But if there's no choice, I guess there's no choice.

Another option would be to avoid Thread storage altogether.

This would probably require serious code changes to the custom memory allocator and the Hash Maps. Fixing this for the `fiobj2str` function might be impossible without API changes as well (requiring the function to accept a self supplied buffer).

janbiedermann:
It seems like it should work out of the box, but doesnt for some reason.
If the tls hasnt been initialized, it tries to initialize it:

BT: C:\Ruby30-x64\bin\ruby_builtin_dlls\libgcc_s_seh-1.dll(_emutls_get_address+0x122) [0x00007ffdc7f1dd62]
https://github.com/gcc-mirror/gcc/blob/7cc2df084b7977653a9b59cbc34a9ad500ae619c/libgcc/emutls.c#L144

But then later fails, when initializing the actual tls object:
BT: C:\Windows\System32\msvcrt.dll(memcpy+0xf) [0x00007ffde25843cf]
https://github.com/gcc-mirror/gcc/blob/7cc2df084b7977653a9b59cbc34a9ad500ae619c/libgcc/emutls.c#L119

boazsegev:
Your debugging skills are far superior to my own! Nicely caught!

I will work on extracting all use cases of TLS from the implementation (at least on Windows if not everywhere), but this is bound to cause some API changes, which means that the changes will only be implemented in facil.io 0.8.x onwards (semantic versioning and all that).

For using the 0.7.x branch in iodine we might consider avoiding the API calls that require TLS (i.e., `fiobj_each`) and choose instead to iterate through the Hash / Array manually.

janbiedermann:
I dont know about __thread on linux, but on Windows using the pthread tls pattern directly seems benefitial, as all those gcc emutls calls can be avoided, which use the pthreads calls internally anyway. The pattern is not too bad:
```
static pthread_key_t each_at_key;
static pthread_once_t each_at_key_once = PTHREAD_ONCE_INIT;
static void init_each_at_key(void) {
  pthread_key_create(&each_at_key, NULL);
  pthread_setspecific(each_at_key, (void *)FIOBJ_INVALID); // initalize to some value whatever
}

/** later on use like this: */
pthread_once(&each_at_key_once, init_each_at_key);
FIOBJ old_each_at_key = (FIOBJ)pthread_getspecific(each_at_key);
```
Its just making __thread explicit, at least on Windows, thats it.
I am in process of implementing that for iodine and will then copy to facil.io 0.7.x, if it actually works.

janbiedermann:
The pthread tls pattern seems to work, but not tested very well yet, as iodine crashes. Why do you use FIX2ULONG here in iodine_rack_io.c:79:
```
static inline FIOBJ get_data(VALUE self) {
  VALUE i = rb_ivar_get(self, io_id);
  return (FIOBJ)FIX2ULONG(i);
}
```
?
FIX2ULONG seems to mangle VALUE causing a crash later on, but not for all VALUEs.
eg: `get_data VALUE: 443418752 FIX2ULONG: 221709376`
Removing FIX2ULONG  makes things work a bit more, till next crash.

janbiedermann:
Well, i am using Ruby 3.0, this maybe had to be this way for Ruby 2.x?

boazsegev:
>  The pthread tls pattern seems to work...

I imagine it would work well... however, I think it doesn't really excuse the use of thread local storage. I think I should be less lazy in my design and find a way to avoid both the performance and portability costs related to thread local storage. For example, this might conflict with the possible use of Coroutines (i.e., Ruby Fibers) where an alternative solution might work seamlessly.

> Why do you use FIX2ULONG [here](https://github.com/boazsegev/iodine/blob/81940f2e4f5b2fd2dbb9580cdc65f07c205f892b/ext/iodine/iodine_rack_io.c#L79-L82)...

TL.DR.: **on Windows you can probably fix the crush by replacing these macros with the `unsigned long long` alternatives: [`ULL2NUM` and `NUM2ULL`](https://github.com/ruby/ruby/blob/d92f09a5eea009fa28cd046e9d0eb698e3d94c5c/include/ruby/internal/arithmetic/long_long.h#L28-L35), sine the type lengths on POSIX and Windows are different**.

There's a foolish reason that actually helps (it's called laziness and optimism), but I believe that a better solution might have been to [use `TypedData_Wrap_Struct`, as done here](https://github.com/boazsegev/iodine/blob/81940f2e4f5b2fd2dbb9580cdc65f07c205f892b/ext/iodine/iodine_connection.c#L70-L155).

The reason is his:

`FIOBJ` are basically just tagged pointer values, which means they fit in an `unsigned long`... which is 8 bytes long on most POSIX systems (on windows it is only 4 bytes long and  we need to use `unsigned long long`, which requires the [`ULL2NUM` and `NUM2ULL` macros](https://github.com/ruby/ruby/blob/d92f09a5eea009fa28cd046e9d0eb698e3d94c5c/include/ruby/internal/arithmetic/long_long.h#L28-L35)).

By storing the FIOBJ data as a number, I avoided writing the code to another container type and avoiding a few allocations.

However, it appears that I may have been optimistic, as Ruby uses a similar pointer tagging approach in order to store some numbers within the 8 byte pointer `VALUE` (which is why the `VALUE` type requires some translation before it can be used as a pointer)... this means that Ruby probably allocates memory to store the FIOBJ variable (which means I saved myself some code writing, but probably didn't save as many allocations as I may have hoped).

janbiedermann:
Thanks for the explanation, seems to work so far in iodine :-)

PR #107, opened 2021-07-06
 Fix for gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04):

janbiedermann:
lib/facil/fio.c: In function ‘fio_start’:
lib/facil/fio.c:3923:1: error: embedding a directive within macro arguments is not portable [-Werror]
 3923 | #if HAVE_OPENSSL


boazsegev (approved):
As much as I dislike the repetition, I think you're right.

PR #108, opened 2021-07-06
Make test run by default, avoid *void function cast error

janbiedermann:


boazsegev:
I like that we can avoid the warning... but... do we really want to avoid the warning?

Wouldn't it be better to have explicit code (i.e. `(void*)(intptr_t)(func)`) rather than miss instances where this comment actually reveals an unintended cast?

I think finding any warnings in the code would be better 🤔?

janbiedermann:
Nono, its not avaoiding the warning, its avoiding the error. The error makes compilation stop. I guess this is since gcc 9 or so with -pedantic, having these casts is a problem. There is no simple solution, even casting will print a warning/error so this just makes it print the warning and compilation succeed. It really needs another solution, like VAARGS or args with the relevant types.

boazsegev:
I think I will need to test alternatives before I disable a warning that might actually indicate an unintended error.

I was pretty sure an explicit "chain" casting using `uintptr_t` can be used to prevent these types of warnings by informing the compiler that this cast is intentional (i.e., `(int (*)(int)(intptr_t)fn`)... but maybe I missed something.

PR #111, opened 2021-07-06
assert alloc

janbiedermann:


lib/facil/http/http_internal.c
```diff
@@ -98,9 +98,11 @@ int http_send_error2(size_t error, intptr_t uuid, http_settings_s *settings) {
     return -1;
   fio_protocol_s *pr = http1_new(uuid, settings, NULL, 0);
   http_s *r = fio_malloc(sizeof(*r));
+  FIO_ASSERT_ALLOC(r);
```
boazsegev, line 101: Sweet :)

lib/facil/http/http_internal.c
```diff
@@ -98,9 +98,11 @@ int http_send_error2(size_t error, intptr_t uuid, http_settings_s *settings) {
...
   FIO_ASSERT(pr, "Couldn't allocate response object for error report.")
   http_s_new(r, (http_fio_protocol_s *)pr, http1_vtable());
   int ret = http_send_error(r, error);
+  free(r);
```
boazsegev, line 105: Was there actually a memory leak?

The object is supposed to be freed after it is used by the HTTP implementation. See: https://github.com/boazsegev/facil.io/blob/abc4ecf72904fc05fe3bc2dab9cde627de265461/lib/facil/http/http1.c#L54-L56

boazsegev (changes requested):

lib/facil/http/http_internal.c
janbiedermann: I see. Yes, there was a leak somewhere, maybe it was the other one.

PR #112, opened 2021-07-06
fix wrong assignment of max_protocol_fd

janbiedermann:


lib/facil/fio.c
```diff
@@ -307,7 +307,7 @@ static inline int fio_clear_fd(intptr_t fd, uint8_t is_open) {
   if (fio_data->max_protocol_fd < fd) {
     fio_data->max_protocol_fd = fd;
   } else {
-    while (fio_data->max_protocol_fd &&
+    while (fio_data->max_protocol_fd > fd &&
```
boazsegev, line 310: Initially I though "Wow! how didn't I see it"... but then I looked again and I think the reading here makes the logic a bit less obvious...

The code assumes that `fio_data->max_protocol_fd > fd` but the code doesn't know if `fd_data(fd).open` is true. `fd` might be now closed and actually higher than what `max_protocol_fd` should be.

This loop stops at either `max_protocol_fd == 0` or `fd_data(fio_data->max_protocol_fd).open` , so there's no risk of an underflow.

However there might be a bug in the if statement above.

lib/facil/fio.c
```diff
@@ -307,7 +307,7 @@ static inline int fio_clear_fd(intptr_t fd, uint8_t is_open) {
   if (fio_data->max_protocol_fd < fd) {
```
boazsegev, line 307: Perhaps this is the line we need to fix by stating `if (is_ope && fio_data->max_protocol_fd < fd)`...?

boazsegev (changes requested):

lib/facil/fio.c
janbiedermann: Alright, i will try

janbiedermann: Tested and changed

boazsegev (approved):
Thanks for updating this one :)

PR #116, opened 2021-07-06
fix locking order causing a deadlock with fio_unsubscribe

janbiedermann:
fio_filter_dup_lock_internal

    locks the collection
    doing stuff, getting channel
    locks the channel
    unlocks the collection

channel gets unlocked in fio_subscribe

fio_unsubscribe

    locks the channel
    doing stuff
    locks the collection
    ...

Lets look at 2 threads running that interleaved by accident:
1 fio_filter_dup_lock_internal
2 fio_unsubscribe
1 lock collection
1 doing stuff getting channel
2 locking channel
2 doing stuff
1 locking channel --> waiting
2 locking collection --> waiting
--> deadlock, busy spinning


boazsegev:
Nice catch! I'm not sure I would have caught on that. It takes an in-depth observation to notice....

....however, I think the solution requires some polishing.

When unsubscribing the last subscriber (evicting the channel), we need to lock the collection while the channel is locked to new subscribers... otherwise a subscriber might be attached to the evicted channel in between unlocking and locking.

On the other hand, while subscribing we need to acquire the lock for the channel while the collection is locked, to prevent the collection from creating another equal channel (for first subscribers) and to prevent users from having their subscriptions attached to a Chanel object that is now being evicted...

So the solution here requires a bit more logic (i.e., re-testing the evicted channel to see if a subscription was added between the release of the channel lock and the acquiring of the collection lock) 🤔

janbiedermann:
Could also change the lock order in fio_unsubscribe maybe?

boazsegev:
I don't see a need to access the collection in `fio_unsubscribe` unless it's the last subscriber (and we need to evict the channel)... locking the collection anyway could add a lot of contention to something that can easily be concurrent (when unsubscribing from different channels).

janbiedermann:
To keep things simple, in case of a deadlock, just throttle the thread a bit, let the subscription finish, and try locking the channel again. Works.

janbiedermann:
i meant to write: ... and try locking the collection again ...

boazsegev:
I think I missed something - this actually works? How?

If the channel is locked, how does throttling prevent a deadlock when the locked collection is attempting to lock the channel and the channel is waiting to lock the connection?

janbiedermann:
Right. So let me rephrase: It _appears_ to work. Sigh.

PR #117, opened 2021-07-06
properly init locks

janbiedermann:


boazsegev (approved):
Yes, definitely!

It is true that `FIO_LOCK_INIT` is currently zero, so the resulting instructions don't actually change. However, this is an implementation detail that might change in the future and it is important to avoid having the rest of the code rely on this "internal" knowledge.

PR #120, opened 2021-07-06
ensure initialization of buf

janbiedermann:


lib/facil/fio.h
```diff
@@ -4339,7 +4339,7 @@ inline FIO_FUNC fio_str_info_s fio_str_write_i(fio_str_s *s, int64_t num) {
   fio_str_info_s i;
   if (!num)
     goto zero;
-  char buf[22];
+  char buf[22] = {0};
```
boazsegev, line 4342: In general I agree, however string to number  and number to string conversions happen so often that one might consider them a critical performance path... so why spend CPU cycles on this? I see no reason to initialize the buffer, as we never test for a `NUL` terminator byte. Am I missing something?

boazsegev:
Following my review comment, if you think I am missing something, please let me know. Otherwise I suggest we skip this PR.

lib/facil/fio.h
janbiedermann: I tried a static analyzer to help me find the show stopper. It was of no help, but suggested this. I thought it made sense. So static analyzers for c are still garbage.

boazsegev: I think it's a good suggestion for the most part... it's safer to make sure memory is always initialized. But sometimes it's more efficient to walk the edge :)

PR #122, opened 2021-07-07
init buf

janbiedermann:


lib/facil/fiobj/fiobject.c
```diff
@@ -102,7 +102,7 @@ size_t __attribute__((weak)) fio_ltoa(char *dest, int64_t num, uint8_t base) {
                            '8', '9', 'A', 'B', 'C', 'D', 'E', 'F'};
 
   size_t len = 0;
-  char buf[48]; /* we only need up to 20 for base 10, but base 3 needs 41... */
+  char buf[48] = {0}; /* we only need up to 20 for base 10, but base 3 needs 41... */
```
boazsegev, line 105: This is a weak version of the function that was made available for when extracting the FIOBJ library out of the facil.io code-base. This code is essentially dead code... so this change doesn't (or shouldn't) do anything...

Generally speaking it's good idea to initialize everything. However, it's also a performance hit that we don't really need since we always explicitly append a `NUL` terminator once we finish writing to the buffer.

i.e.: https://github.com/boazsegev/facil.io/blob/a6569413ef3d25fe5e225d8ffa655c754bf003f7/lib/facil/fiobj/fiobject.c#L155

boazsegev: Unless there's a compelling reason you would add following my comment, I would suggest that we cancel the pull request...?

janbiedermann: No compeeling reason.

PR #125, opened 2021-07-07
assert alloc

janbiedermann:


lib/facil/http/websockets.c
```diff
@@ -81,8 +82,9 @@ struct buffer_s resize_ws_buffer(ws_s *owner, struct buffer_s buff) {
     free_ws_buffer(owner, buff);
     buff.data = NULL;
```
boazsegev, line 83: since `tmp` is known to be `NULL`, it might be easier to avoid the `else` clause and remove the `buff.data = NULL` line (so `buff.data = tmp` always executes.

This might be meaningless, especially on a good compiler, but on a terrible compiler it could improve code cache locality on the CPU by producing less instructions and having a shorter branch.

lib/facil/http/websockets.c
```diff
@@ -71,6 +71,7 @@ struct buffer_s create_ws_buffer(ws_s *owner) {
   struct buffer_s buff;
   buff.size = WS_INITIAL_BUFFER_SIZE;
   buff.data = malloc(buff.size);
+  FIO_ASSERT_ALLOC(buff.data);
```
boazsegev, line 74: Nice catch 👍🏻

lib/facil/http/websockets.c
```diff
@@ -81,8 +82,9 @@ struct buffer_s resize_ws_buffer(ws_s *owner, struct buffer_s buff) {
     free_ws_buffer(owner, buff);
     buff.data = NULL;
     buff.size = 0;
+  } else {
+    buff.data = tmp;
```
boazsegev, line 86: Should always execute.

lib/facil/http/websockets.c
```diff
@@ -330,7 +334,6 @@ static void destroy_ws(ws_s *ws) {
 void websocket_attach(intptr_t uuid, http_settings_s *http_settings,
                       websocket_settings_s *args, void *data, size_t length) {
   ws_s *ws = new_websocket(uuid);
-  FIO_ASSERT_ALLOC(ws);
```
boazsegev, line 333: True! 👍🏻

boazsegev (changes requested):
Hi! Thanks for pointing these out. There's just once small detail that is almost a style (if not a performance) consideration, have a look at the comments.

lib/facil/http/websockets.c
janbiedermann: Sure. Ill change it.

janbiedermann: changed

boazsegev (approved):
Thanks :)

